PR #269, opened 2022-08-25
Added ability to change the background fill color of a shape

femo1de:
This is a super early implementation but it gets the job done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## Version 0.33.0 - Unreleased
+- Added `IParagraph.AddPortion(string text)` to add a new text portion in paragraph [#297](https://github.com/ShapeCrawler/ShapeCrawler/issues/297)
+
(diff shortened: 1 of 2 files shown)


adamshakhabov:
Hello @femo1de. Thank you for your PR!

There is `IAutoShape.ShapeFill` class for shape fill behavior. Would be good to add your changes in this level.

Then you can name it `ShapeFill.SetSolidColor()`

adamshakhabov:
@femo1de, did you have time to update PR?

femo1de:
Hi Adam,

sorry I havent had the time yet. I tried adding it to IShapeFill but I found that this would require a lot of refactoring since ShapeFill is used for several types of shapes including images, not all of which who support background colors. Also it doesnt store a reference to the xml element of the ppt. I agree this would have been a better place to add this functionality but unfortunately I havent had the time yet to make these changes.
________________________________

adamshakhabov:
Thank you for the updates.

PR #296, opened 2022-09-13
IParagraph.AddText implementation

sergsuhanov:
Adding text creates IPortion entries so you can change the font and color.

```
var newPar = paragraphs.Add();
newPar.Text = "Text";
newPar.AddText("Text2");

var last = newPar.Portions.Last();
var font = last.Font;
font.IsItalic = true;
font.Size = 18;
```


ShapeCrawler/AutoShapes/IParagraph.cs
```diff
@@ -9,6 +9,11 @@ namespace ShapeCrawler
     /// </summary>
     public interface IParagraph
     {
+        /// <summary>
+        /// Add text to paragraph.
+        /// </summary>
+        void AddText(string text);
```
adamshakhabov, line 15: @sergsuhanov, thank you for the PR.

What about moving this to `IPortionCollection` level? I think something like `newPr.Portions.Add("Text2")` would be more consistent.

sergsuhanov: hi, maybe
IPortion maps to A.Run and when we add a text separator we use A.Break.

`var lastElement = this.AParagraph.Where(p => p is A.Run || p is A.Break).Last(); // for inset after this`
I think that at the paragraph level, working with text is more logical.

adamshakhabov: Got it. Then what about rename `AddText()` -> `AddPortion()`

sergsuhanov: Ok, sure

adamshakhabov (approved):

PR #302, opened 2022-09-16
Added helper to return all textboxes on a slide

femo1de:
This property will return all textboxes that are on a slide, including those that are inside Cells which are nested in a table. This is useful for scenarios where you try to find specific strings and replace them, regardless of whether they are inside a textbox, autoshape or in a tables cell.


ShapeCrawler/PowerPoint/ISlide.cs
```diff
@@ -41,6 +43,11 @@ public interface ISlide
         
         IShapeCollection Shapes { get; }
 
+        /// <summary>
+        /// Gets a list of all textboxes on that slide, including those in tables
+        /// </summary>
+        public IList<ITextBox> Textboxes { get; }
```
adamshakhabov, line 49: @femo1de, thank you for the PR.

I think the _method_ concept, eg. `GetAllTextboxes()` or `GetTextBoxList()`, would be more suitable here. Textboxes is not _direct property_ of the slide from user's perspective.

adamshakhabov (changes requested):
Please add unit test.

ShapeCrawler/PowerPoint/ISlide.cs
adamshakhabov: @femo1de, did you have chance to resolve the review comments?

femo1de: @adamshakhabov they should all be solved now

adamshakhabov: Pull Request contains a lot of other changes:

image

Looks like the conflict resolving while `merge`/`rebase` operation was made incorrectly. For example, the "Files changed" tab says, the following changes are your, which is not true:

image

femo1de:
I've created a new PR, I think I screwed up somewhere with merging here 😄

PR #310, opened 2022-09-20
add editable property SCBullet and add property Underline for IFont

sergsuhanov:


adamshakhabov (changes requested):
@sergsuhanov, thank you for the PR.

Please, add unit test.

adamshakhabov (approved):

PR #324, opened 2022-09-27
Clone / Remove rows in a table

femo1de:
## Proposed Changes
Changes table implementation to allow users to:
- Clone a row in memory 
`public SCTableRow SCTableRow.CloneRow() `    -> this will return a clone of the current row
- Append a row to a table
`public void ITable.AppendRow(TableRow)`   -> row will added to the bottom of a table
- remove a row by index
`public void ITable.RemoveAt(int)`   -> On a table, user can remove a row by its index

## Check List

- [x] Added unit-test


ShapeCrawler.Tests/Properties/Resources.resx
```diff
@@ -238,4 +238,7 @@
   <data name="_040_video" type="System.Resources.ResXFileRef, System.Windows.Forms">
     <value>..\Resource\040_video.pptx;System.Byte[], mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </data>
+  <data name="_041_tableOnly" type="System.Resources.ResXFileRef, System.Windows.Forms">
+    <value>..\Resource\041_tableOnly.pptx;System.Byte[], mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
```
adamshakhabov, line 242: Could you please rename `Resource\041_tableOnly.pptx` on  `Resource\tables\tables-case001.pptx`.

There is an idea of no longer using `Resource\{xxx}.pptx` pattern for the naming of test file. Would be better to have folder for each shape type. I believe the navigation is better in this way.

adamshakhabov (changes requested):

ShapeCrawler.Tests/Properties/Resources.resx
femo1de: done

adamshakhabov (approved):

PR #500, opened 2023-04-23
Remove table column

freshe:
Ignore this if it's already possible, or if you plan a different approach. I couldn't find a way to remove a table column.

I'm not too familiar with the OpenXML specs and schema, but this patch works in our environment. I looked at the MergeCells method and used the same approach. Seems to work with merged cells as well.

---

## PR-Codex overview
This PR adds a new method to remove a table column in ShapeCrawler. 

### Detailed summary
- Added `RemoveColumn` method to `ITable` interface
- Implemented `RemoveColumn` method in `SCTable` class
- Added a new unit test to test the `RemoveColumn` method

### DIFF
--- a/src/ShapeCrawler/Tables/ITable.cs
+++ b/src/ShapeCrawler/Tables/ITable.cs
@@ -41,7 +41,7 @@ public interface ITable : IShape
     void MergeCells(ICell cell1, ICell cell2);
 
     /// <summary>
-    ///     Deletes a column at specified index
+    ///     Removes a column at specified index.
     /// </summary>
     void RemoveColumn(int columnIndex);
 }
--- a/src/ShapeCrawler/Tables/ITable.cs
+++ b/src/ShapeCrawler/Tables/ITable.cs
@@ -39,6 +39,11 @@ public interface ITable : IShape
     ///     Merge neighbor cells.
     /// </summary>
     void MergeCells(ICell cell1, ICell cell2);
+
+    /// <summary>
+    ///     Deletes a column at specified index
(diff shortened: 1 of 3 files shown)


src/ShapeCrawler/Tables/ITable.cs
```diff
@@ -39,6 +39,11 @@ public interface ITable : IShape
     ///     Merge neighbor cells.
     /// </summary>
     void MergeCells(ICell cell1, ICell cell2);
+
+    /// <summary>
+    ///     Deletes a column at specified index
```
adamshakhabov, line 44: The build fails here with "Documentation text should end with a period". Please build the **Release** configuration on your local machine to see all code style issues.

src/ShapeCrawler/Tables/ITable.cs
```diff
@@ -39,6 +39,11 @@ public interface ITable : IShape
...
+    /// </summary>
+    void RemoveColumn(int columnIndex);
```
adamshakhabov, line 46: It is ok! But I think later we should change this to `ITable.Columns.RemoveAt(int index)`.

adamshakhabov (changes requested):
@freshe, thank you for the Pull Request.

I've left a few comments.

adamshakhabov:
Correct. We should create a new interface `IColumnCollection` and the new class `SCColumnCollection` that inherites this.

As I said before, we will do this in the scope of another feature.

adamshakhabov (approved):

PR #505, opened 2023-05-08
Clear/Remove text highlight doesn't work

huchim:
Hi!

I have problems using text highlight feature.

I created two portions, but I want to set hightlight only to first portion, and clear or remove it in the next portion. My code use `portion2.TextHighlightColor = null` to "clear" hightlight.

You may reproduce the problem with this code:

```cs
using ShapeCrawler;

// Creates a new presentation with one shape
var pres = SCPresentation.Create();
var shapes = pres.Slides[0].Shapes;
var rectangle1 = shapes.AddRectangle(10, 20, 200, 100);

// We add two portions
rectangle1.TextFrame!.Paragraphs.First().AddPortion("FOO");
rectangle1.TextFrame.Paragraphs.First().AddPortion("BAR");
var portion1 = rectangle1.TextFrame.Paragraphs.First().Portions[0];
var portion2 = rectangle1.TextFrame.Paragraphs.First().Portions[1];

// Text: white, highlight: black
portion1.Font.ColorFormat.SetColorByHex("FFFFFF");
portion1.TextHighlightColor = "FF0000";

// Text: black, highlight: none
portion2.Font.ColorFormat.SetColorByHex("000000");
portion2.TextHighlightColor = null;

// Save file in %temp% directory
var fileName = Path.GetTempPath() + Guid.NewGuid().ToString() + ".pptx";
pres.SaveAs(fileName);

Console.WriteLine("Output: {0}", fileName);
```

Output:

imagen

As you can see, next portion have a "black" highlight. 

For some reason, "null" value is accepted and your codes try to add a new hightlight node with null value:

https://github.com/ShapeCrawler/ShapeCrawler/blob/21ae1a6ffb089c150c711d269fb77bbf76bbf2c7/src/ShapeCrawler/Extensions/RunPropertiesExtensions.cs#LL18C14-L18C14

If you add:

```cs
// Don't add a new node
if (hex is null)
{
  return;
}
```

This PR is for accepting "nulls" in `AddAHighlight` as a way to "clear" text hightlighting.

Thank you so much 👍


adamshakhabov:
Hello @huchim,

Thank you for the Pull Request👍

I think "Passing `null` to the method" doesn't look good. What about adding a new method `IPortion.SetTextHighlight(SCColor color)` and new enum `SCColor`:

```
public enum SCColor
{
    Transparent, // to remove highlighting: IPortion.SetTextHighlightColor(SCColor.Transparent)
    
    //...
}
```

This way, we can _later_ extend `SCColor` with regular colors like `SCColor.White`, `SCColor.Black`, etc. And maybe `IPortionTextHighlightColor` will be removed.

huchim:
Thank you for your feedback!

I created a new class `SCColor` (and some tests) with a few methods: 

- `SCColor(System.Drawing.Color color)`: Creates a new instance using system color.
- `SCColor(int red, int green, int blue)`: Creates a new instance using a RGB values.
- `SCColor(int red, int green, int blue, float alpha)`: Creates a new instance using RGBA values.
- `TryGetColorFromHex(string hext, out SCColor result)`: Try to parse an hex value from string.

Class has two internal properties: `IsTransparent` (Alpha = 0) and `IsSolid` (Alpha1 = 255) in order to detect how to add color. 

I added one overloading of extension method `RunPropertiesExtensions.AddAHighlight(SCColor hex` (to keep original method) and `ToRgbColorModelHex()` to cast `SCColor` to `A.RgbColorModelHex`

😄  *ToRgbColorModelHex()*
```cs
internal static A.RgbColorModelHex ToRgbColorModelHex(this SCColor color)
{
    // ...
    var alpha = new A.Alpha
    {
        Val = (Int32Value)(100000f * (color.Alpha / 255))
    };

    model.AddChild(alpha);

   return model;
}
```

Following code creates a new file:

```cs
using ShapeCrawler;

// Creates a new presentation with one shape
var pres = SCPresentation.Create();
var shapes = pres.Slides[0].Shapes;
var rectangle1 = shapes.AddRectangle(10, 20, 600, 600);

// We add two portions
rectangle1.TextFrame!.Paragraphs.First().AddPortion("FOO");
rectangle1.TextFrame.Paragraphs.First().AddPortion("BAR");
rectangle1.TextFrame.Paragraphs.First().AddPortion("KAA");
var portion1 = rectangle1.TextFrame.Paragraphs.First().Portions[0];
var portion2 = rectangle1.TextFrame.Paragraphs.First().Portions[1];
var portion3 = rectangle1.TextFrame.Paragraphs.First().Portions[2];

// Text: white, highlight: black
portion1.Font.ColorFormat.SetColorByHex("FFFFFF");
portion1.TextHighlightColor = "FF0000";

// Text: black, highlight: none
portion2.Font.ColorFormat.SetColorByHex("000000");
portion2.SetTextHighlight(ShapeCrawler.Drawing.SCColor.Transparent);

// Text: black, highlight: (alpha = 125)
portion3.Font.ColorFormat.SetColorByHex("000000");
portion3.SetTextHighlight(new ShapeCrawler.Drawing.SCColor(35, 23, 44, 125));

var fileName = "hola.pptx";
pres.SaveAs(fileName);
```
*Output*: (I updated file using Powerpoint to add a shape behind:

imagen

This the XML:

```xml
<a:r>
  <a:rPr lang="en-US">
    <a:solidFill>
      <a:srgbClr val="000000"/>
    </a:solidFill>
    <a:highlight>
      <a:srgbClr val="23172C">
        <a:alpha val="49019"/>
      </a:srgbClr>
    </a:highlight>
  </a:rPr>
  <a:t>KAA</a:t>
</a:r>
```

I don't know how to set opacity, I assumed that `100000` is the max value, `Val = (Int32Value)(100000f * (color.Alpha / 255))` is used in `A.Alpha` node.

What do you think? maybe this `SetTextHighlight` is in conflict with `TextHighlightColor` property.

Thanks!

src/ShapeCrawler/Drawing/SCColor.cs
```diff
@@ -0,0 +1,206 @@
+﻿using System;
+using System.Drawing;
```
adamshakhabov, line 2: Unfortunately, we can't use `System.Drawing` namespace because it [only supports Windows starting from .NET 6](https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/system-drawing-common-windows-only). That's why we recently switched to [SkiaSharp](https://github.com/mono/SkiaSharp).

adamshakhabov (changes requested):

adamshakhabov:
> What do you think? maybe this SetTextHighlight is in conflict with TextHighlightColor property.

Yes, since we now have `SCColor.TryGetColorFromHex()`, then `TextHighlightColor` can be removed now or later.

src/ShapeCrawler/Drawing/SCColor.cs
huchim: I removed System.Drawing references.

huchim:
This new methods are included in this PR.

```cs
public interface IPortion
{
    /// <summary>
    /// Set hightlight.
    /// </summary>
    /// <param name="color">Color value.</param>
    void SetTextHighlight(SCColor color);

    /// <summary>
    /// Gets highlight color.
    /// </summary>
    /// <returns>Returns <see langword="null" /> if  the text doesn't have a highlight defined.</returns>
    SCColor? GetTextHighlight();
}
```

:1st_place_medal: Expected behavior of `GetTextHighlight()`:

- Returns a `SCColor`.
- Returns `null` is `RgbColorModelHex` doesn't exists or his value is null.
- Throws an exception if value is invalid (`SCColor.TryGetColorFromHex()`)
- Calculate alpha value.

:1st_place_medal: Expected behavior of `SetTextHighlight(SCColor color)`:

- Remove `A.Highlight` child if `color` is transparent.
- Add a `RgbColorModelHex`.
  - Creates a `A.Alpha` child if color has transparency.

:1st_place_medal: Expected behavior of `SCColor`

- Initializes a new color using `string`, RGB or RGBA values.
- Try to parse a RGB(A) color (`TryGetColorFromHex()`)
- Set "some" default colors (Transparent, Black and White)
- Transparent value is = `new (0, 0, 0, 0)`

Maybe we can mark as obsolete `TextHighlightColor` property or change property type to `SCColor`.

:grin: Thanks!

huchim:
PR is on board! :ship:

PR #537, opened 2023-06-30
Use `AutoShapeCreator` when add a shape to an slide

### LINKED ISSUE #536: Unable to cast object of type 'ShapeCrawler.SCAutoShape' to type 'ShapeCrawler.ITable'
There's an error when trying to add a table to an slide.

Shape collection is returning a `SCTable` (`sourceTable`) but method `Add` is returning a `SCAutoShape`.

```cs
// SCTable
var sourceTable = ComponentsSlide.Shapes.First(c => c.Name == _options.TableName);
// SCAutoShape
var targetTable = (ITable)record.Slide.Shapes.Add(sourceTable);
// Unable to cast object of type 'ShapeCrawler.SCAutoShape' to type 'ShapeCrawler.ITable'
```

These classes inherits from `SCShape`:

- `SCChart` and `SCOLEObject` are not supported in the [529](https://github.com/ShapeCrawler/ShapeCrawler/pull/535)
- `SCGroupShape`: I don't how to return a new class of this type.
- `SCPicture`, `SCImage`, `SCTable`: I can create a factory for this types.
- `SCAutoShape`: This is the default.

I will work on a new PR 😄

huchim:
This PR tries to fix #536 

The `ShapeCollection` constructor creates an instance of `AutoShapeCreator` that traverse nodes to creates the correct `IShape` implementation. This PR makes that instance a private variable so `ShapeCollection.Add(IShape)` can use it.

It works fine in rectangles, lines, shapes groups, textboxes, tables, but it fails on charts, pictures, audio, video and ole object and I don't know how to support that types of shapes.

---

### Detailed summary
- Added a unit test to verify that adding a table shape to a shape collection returns a new table.
- Added an `autoShapeCreator` field to the `ShapeCollection` class.
- Modified the `Add` method in the `ShapeCollection` class to handle different shape types.
- Added a `GetShape` method to the `ShapeCollection` class to create and return a shape based on its type.


adamshakhabov (changes requested):
Please add a unit test. It's enough if you add one test that checks one of the new cases, like checking that the response of `.Slide.Shapes.Add(sourceTable)` castable to `ITable`.

huchim:
okas... I will do it right now...

adamshakhabov (approved):

Issue #539, opened 2023-06-30
Add Group Shape with IShapeCollection.Add()


adamshakhabov:
Add the ability to add `IGroupShape` by `IShapeCollection.Add()`.

PR #542, opened 2023-07-04
Add support for add picture to shape collection

huchim:
`AutoShapeCreator` works fine, but pictures requires to add the image part from source slide to the target slide.

```cs
if (newShape is SCPicture pic)
{
    pic.CopyParts((SlideStructure)scShape.ParentSlideStructureOf.Value);
}
```

This is my first approach to solve the problem, `CopyParts` copy the relationship to the target slide and update the embed value (a new relationship id is created)

```cs
var sImagePart = sSlidePart.GetPartById(relId);
var c = slidePart.AddPart(sImagePart, imgPartRId);
this.blipEmbed.Value = imgPartRId;
```

wht do you think?

---

### Focus of this PR:
This PR focuses on improving the shape collection and picture handling in the ShapeCrawler library.

### Detailed summary:
- Removed unsupported shape types from the shape collection.
- Added a method to copy required parts from a source slide to a picture shape.
- Updated the picture shape class to handle copying of parts from a source slide.

### DIFF
--- a/src/ShapeCrawler/Collections/IShapeCollection.cs
+++ b/src/ShapeCrawler/Collections/IShapeCollection.cs
@@ -158,7 +158,7 @@ internal sealed class ShapeCollection : IShapeCollection
         // Chart (<p:graphicFrame /> http://schemas.openxmlformats.org/drawingml/2006/chart) are not in the shape collection, data is referenced.
         // Object (<p:graphicFrame /> http://schemas.openxmlformats.org/presentationml/2006/ole) are not in the shape collection, data is referenced.
         // Alternate content(<mc:AlternateContent /> http://schemas.openxmlformats.org/officeDocument/2006/math"> are not in the shape collection, data is referenced.
-        if (shape is not SCShape scShape || shape is IOLEObject || shape is IChart || shape is IPicture || shape is IAudioShape || shape is IVideoShape)
+        if (shape is not SCShape scShape || shape is IOLEObject || shape is IChart || shape is IAudioShape || shape is IVideoShape)
         {
             throw new SCException($"{shape.GetType().Name} is not supported.");
         }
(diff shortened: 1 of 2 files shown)


huchim:
#539

adamshakhabov:
Have you checked your solution for adding a picture from an _external_ presentation?

The high-level steps for adding a picture from an external presentation would be:

1) Create a new `ImagePart` in target slide.
2) Copy the `ImagePart` stream from the source picture into the `ImagePart` (step 1).
3) Copy the shape of source picture to the target slide.
4) Use the id the `ImagePart` (step 1) as the image source for the added picture (step 3).

huchim:
I made the changes, but in the step 4, i created a new "id" in order to avoid other id in the target slide.

`SCPicture.CopyParts()` is intended to be used only for `ShapeCollection.Add()` so is marked as `internal`.

The image extension is ".bin" and this is a related issue [New ImagePart is created as .bin instead of real extension](https://github.com/dotnet/Open-XML-SDK/issues/1305).

adamshakhabov:
Hmm, you are right. For example, I've checked the output of the test [ShapeFillTests.SetPicture_updates_fill_with_specified_picture_image_When_shape_is_Not_filled](https://github.com/ShapeCrawler/ShapeCrawler/blob/master/test/ShapeCrawler.Tests.Unit/ShapeFillTests.cs#L36) => is broken duo to the .bin issue😢

We need to keep an eye on when the bug gets fixed in the underlying Open XML SDK or find some workaround. I have no idea yet.

huchim:
The method `XmlPartContainer.AddExtendedPart(string relationshipType, string contentType, string targetExt)` has an extension parameter, but maybe is better waits for a fix.

In the meantime, PowerPoint is able to open the file and display the picture.

Issue #567, opened 2023-08-08
Read Theme Color details


ashahabov:
Add APIs to read Theme Color details, such as:

- Alpha (transparency) value of the color (`a:alpha`)
- Luminance modifier for the color (`a:lumMod`)

XML samples

```xml
<a:solidFill>
  <a:schemeClr val="accent2">
    <a:lumMod val="75000"/>
  </a:schemeClr>
</a:solidFill>

<a:solidFill>
  <a:schemeClr val="accent2">
    <a:alpha val="70000"/>
  </a:schemeClr>
</a:solidFill>
```

PR #570, opened 2023-08-09
Enabled dynamically fetching scheme colors without hardcoding maps or color types. Also fixed so SCShapeFill would correctly get a scheme color.

rickszyr:
### Detailed summary
- `ISlideMaster` now has an overridden property `SlideMaster`.
- `ISlideLayout` now has a property `Type`.
- `IShapeCollection` now has an overridden property `Shapes`.
- `IShapeFill` now has additional properties `AlphaPercentage`, `LuminanceModulationPercentage`, and `LuminanceOffsetPercentage`.
- Added new tests for `ShapeFill` properties.

> The following files were skipped due to too many changes: `src/ShapeCrawler/Drawing/HexParser.cs`


rickszyr:
Fixes #567

src/ShapeCrawler/Drawing/IShapeFill.cs
```diff
@@ -23,6 +23,21 @@ public interface IShapeFill
...
+
+    /// <summary>
+    /// Gets the Luminance Modulation Percentage of the fill color.
+    /// </summary>
+    public double? LuminanceModulationPercentage { get; }
+    
+    /// <summary>
+    /// Gets the Luminance Offset Percentage of the fill color.
+    /// </summary>
+    public double? LuminanceOffsetPercentage { get; }
```
adamshakhabov, line 39: I would suggest avoiding `null` when it's _possible_. Can we return 100% for Alpha Percentage, 100% for Luminance Modulation Percentage, and 0% for Luminance Offset Percentage instead of `null`?

rickszyr: We could but it's less of a reflection of what happens in the Powerpoint file. These fields literally don't exist in the elements when they are null. If we're aiming towards an object representation of the xml, I'd leave it as is, otherwise if we're trying to simplify and save time, we can assume those values: 100% Alpha, 100% Luminance Modulation Percentage and 0% Luminance Offset Percentage

rickszyr: I fixed according to this suggestion and also added alpha for when it's a solid color fill

adamshakhabov: Thank you👍

I realize it might be wrong, but I always hope that users will make a bug report with specific case.

PR #630, opened 2024-02-27
Fix `Memory stream is not expandable`

BiDuang:
## PR-Codex overview
This PR improves the initialization process in `PresentationCore` by allowing the constructor to accept both byte array and stream inputs.

### Detailed summary
- Updated `PresentationCore` constructor to accept both byte array and stream inputs for initializing a PowerPoint presentation.
- Improved memory handling by creating a `MemoryStream` object for byte array input.
- Updated initialization of `sdkPresDocument` using the provided stream.

### DIFF
--- a/src/ShapeCrawler/Presentations/PresentationCore.cs
+++ b/src/ShapeCrawler/Presentations/PresentationCore.cs
@@ -18,8 +18,17 @@ internal sealed class PresentationCore
     private readonly SlideSize slideSize;
 
     internal PresentationCore(byte[] bytes)
-        : this(new MemoryStream(bytes))
     {
+        var stream = new MemoryStream();
+        stream.Write(bytes, 0, bytes.Length);
+        stream.Position = 0;
+        this.sdkPresDocument = PresentationDocument.Open(stream, true);
+        var sdkMasterParts = this.sdkPresDocument.PresentationPart!.SlideMasterParts;
+        this.SlideMasters = new SlideMasterCollection(sdkMasterParts);
+        this.Sections = new Sections(this.sdkPresDocument);
+        this.Slides = new Slides(this.sdkPresDocument.PresentationPart!.SlideParts);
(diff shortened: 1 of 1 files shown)


ashahabov:
@BiDuang, thank you for the PR! Please add a Unit test for this fix.

BiDuang:
> @BiDuang, thank you for the PR! Please add a Unit test for this fix.

Well, this feature doesn't required a unit test because it can be easily check by a debugger.

For example, you create a new project (just like #617 code) and add a breakpoint, you will find that the stream's `expandable` now become `true`.

Init a MemoryStream with constructor will make the stream become **NOT** expandable, but create it with no parameter and write the bytes into the stream can slove this problem.

ashahabov:
I mean, if presentation saving throws an exception, then the test can look like this.

```c#
[Test]
public void Save_should_not_throw_exception()
{
    // Arrange
    var pres = new Presentation(StreamOf("some-file.pptx"));
    
    //...

    // Act
    //...
    var act = () => pres.Save();

    // Assert
    act.Should().NotThrow();
}
```

BiDuang:
> I mean, if presentation saving throws an exception, then the test can look like this.
> 
> ```cs
> [Test]
> public void Save_should_not_throw_exception()
> {
>     // Arrange
>     var pres = new Presentation(StreamOf("some-file.pptx"));
>     
>     //...
> 
>     // Act
>     //...
>     var act = () => pres.Save();
> 
>     // Assert
>     act.Should().NotThrow();
> }
> ```

I tried to add some tests and compare the results before and after, but they all passed. I'm confused, but I suspect that the issue (#617) might be caused by the constructor.

Reference: https://learn.microsoft.com/en-us/dotnet/api/system.io.memorystream?view=net-8.0#constructors

BiDuang:
Fixed in `17c9e71a88bbc1d7d86d08a780504333bcddfa6c`

PR #668, opened 2024-05-03
Shapecrawler AddNofill

brunnobb:
## PR-Codex overview
The focus of this PR is to introduce a new method `SetNoFill` to remove fills from shapes and tables in the ShapeCrawler library.

### Detailed summary
- Added `SetNoFill` method to remove fills from shapes and tables.
- Modified existing methods to support `SetNoFill`.
- Added tests for setting no fill on shapes and tables.

### DIFF
--- a/src/ShapeCrawler/Drawing/IShapeFill.cs
+++ b/src/ShapeCrawler/Drawing/IShapeFill.cs
@@ -49,7 +49,7 @@ public interface IShapeFill
     void SetColor(string hex);
 
     /// <summary>
-    ///     Removes Fills from the shape. And set a A.NoFill element.
+    ///     Removes Fills from the shape.
     /// </summary>
     void SetNoFill();
 }
--- a/src/ShapeCrawler/Drawing/IShapeFill.cs
+++ b/src/ShapeCrawler/Drawing/IShapeFill.cs
@@ -47,4 +47,9 @@ public interface IShapeFill
     ///     Fills the shape with solid color in hexadecimal representation.
     /// </summary>
     void SetColor(string hex);
+
+    /// <summary>
+    ///     Removes Fills from the shape. And set a A.NoFill element.
+    /// </summary>
+    void SetNoFill();
 }
--- a/test/ShapeCrawler.Tests.Unit/ShapeFillTests.cs
+++ b/test/ShapeCrawler.Tests.Unit/ShapeFillTests.cs
@@ -191,7 +191,7 @@ public class ShapeFillTests : SCTest
         shapeFill.SetNoFill();
 
         // Assert
-        Type_returns_fill_type(shape, FillType.NoFill);
+        shape.Fill.Type.Should().Be(FillType.NoFill);
         pres.Validate();
     }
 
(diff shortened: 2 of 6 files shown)


test/ShapeCrawler.Tests.Unit/ShapeFillTests.cs
```diff
@@ -171,6 +171,30 @@ public void SetColor_sets_solid_color(string file, int slideNumber, string shape
...
+        // Arrange
+        var pres = new Presentation(StreamOf(file));
+        var shape = pres.Slides[slideNumber - 1].Shapes.GetByName(shapeName);
+        var shapeFill = shape.Fill;
+
+        // Act
+        shapeFill.SetNoFill();
+
+        // Assert
+        Type_returns_fill_type(shape, FillType.NoFill);
```
ashahabov, line 194: @brunnobb, thank you for the PR.

I believe that calling a unit test inside another unit test is not good idea. For example, it leads to Loss Of Isolation: Cascading failures if one test breaks, affecting others that depend on it.

Could you please change such assertions. I think like `shape.Fill.Type.Should().Be(FillType.NoFill);` should be enough.

ashahabov (changes requested):

src/ShapeCrawler/Drawing/IShapeFill.cs
```diff
@@ -47,4 +47,9 @@ public interface IShapeFill
     ///     Fills the shape with solid color in hexadecimal representation.
     /// </summary>
     void SetColor(string hex);
+
+    /// <summary>
+    ///     Removes Fills from the shape. And set a A.NoFill element.
```
ashahabov, line 52: What about just "Removes Fill"?

test/ShapeCrawler.Tests.Unit/ShapeFillTests.cs
brunnobb: Seems good.

I still have a few changes this week, and that will require another PR.
I`ll go back and work on this after the end of our sprint here and submit the changes, as we are currently running on a fork

src/ShapeCrawler/Drawing/IShapeFill.cs
brunnobb: No problem.

Your idea is for the users to not have to know or care about underlying OXML right?

ashahabov: Right. The eventual goal is to have an user friendly Object Model that covers the whole functionalities without having to think about underlying OOXML.

jcoliz:
@brunnobb Love to have this! Just now ran into needed it, not finding it, and voila here is a PR. yay.

test/ShapeCrawler.Tests.Unit/ShapeFillTests.cs
brunnobb: New commit with this change is done

brunnobb:
I am also adding a text Wrapping setter in a new PR after this one is approved.
functionality is ok, just need some unit tests,

jcoliz:
Hey @brunnobb thanks for this, I am using it now. Good stuff. How about `SetNoOutline()` next?? 😁

PR #675, opened 2024-05-12
Add SVG Picture

### LINKED ISSUE #350: Add SVG
Add the `IShapeCollection.AddNewPicture(int xCoordinate,  int yCoordinate, Stream svgContent)` method to able to add SVG picture:

```c#
using var pres = SCPresentation.Open("some.pptx");
var svgContent = new MemoryStream(File.ReadAllBytes("some.svg")); 
var shapeCollection = pres.Slides[0].Shapes;

shapeCollection.AddNewPicture(100, 100, svgStream);
```

jcoliz:
This extends `SlideShapes.AddPicture()` to allow adding SVG pictures.

It does not seem to matter what size the A.Blip gets rasterized at. I've experimented with very small resolutions, but when I load the file in PowerPoint, a proper-looking image is always shown. Ergo, it seems to be fine to simply for the user to add the svg picture and then resize in the usual way by changing the `Height` and `Width` properties. Perhaps a dedicated "resize" API is not needed.

In addition to the included unit tests, I have done some lightweight use-case testing. On Monday, I'll put this into my at-work workflow, which loads hundreds of images from many sources, and we'll see how it does there.

Fixes #350

---

## PR-Codex overview
This PR adds support for adding SVG images to PowerPoint slides and ensures proper handling of SVG content.

### Detailed summary
- Added support for adding SVG images to PowerPoint slides
- Implemented methods to handle SVG content in PowerPoint slides
- Updated project dependencies for SVG support

### DIFF
--- a/src/ShapeCrawler/ShapeCollection/SlideShapes.cs
+++ b/src/ShapeCrawler/ShapeCollection/SlideShapes.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Drawing.Imaging;
 using System.IO;
 using System.Linq;
 using System.Reflection;
@@ -13,7 +14,6 @@ using ShapeCrawler.Services;
 using ShapeCrawler.Shared;
 using SkiaSharp;
 using Svg;
-using System.Drawing.Imaging;
 using A = DocumentFormat.OpenXml.Drawing;
 using A14 = DocumentFormat.OpenXml.Office2010.Drawing;
 using A16 = DocumentFormat.OpenXml.Office2016.Drawing;
--- a/src/ShapeCrawler/ShapeCollection/SlideShapes.cs
+++ b/src/ShapeCrawler/ShapeCollection/SlideShapes.cs
@@ -12,13 +12,13 @@ using ShapeCrawler.Extensions;
 using ShapeCrawler.Services;
 using ShapeCrawler.Shared;
 using SkiaSharp;
-using A = DocumentFormat.OpenXml.Drawing;
-using P = DocumentFormat.OpenXml.Presentation;
-using Position = ShapeCrawler.Positions.Position;
-using A16 = DocumentFormat.OpenXml.Office2016.Drawing;
-using A14 = DocumentFormat.OpenXml.Office2010.Drawing;
 using Svg;
 using System.Drawing.Imaging;
+using A = DocumentFormat.OpenXml.Drawing;
+using A14 = DocumentFormat.OpenXml.Office2010.Drawing;
+using A16 = DocumentFormat.OpenXml.Office2016.Drawing;
+using P = DocumentFormat.OpenXml.Presentation;
+using Position = ShapeCrawler.Positions.Position;
 
 namespace ShapeCrawler.ShapeCollection;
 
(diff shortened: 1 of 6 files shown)


jcoliz:
Last two commits above were done to pass sonarcloud. While I agree with its feedback, and think they are good changes, this now leads to the new code being dissimilar from the existing code. Up to you which is more important: Keeping sonarcloud happy, or congruence with existing code.

I suppose another option would be to make these changes to the existing code as well, so now sonarcloud is happy, code is improved, AND old/new code matches nicely.

src/ShapeCrawler/ShapeCollection/SlideShapes.cs
```diff
@@ -121,18 +125,39 @@ public void AddPicture(Stream imageStream)
...
+            // Not a bitmap, let's try it as an SVG
+            Svg.SvgDocument? doc = null;
+            try
+            {
+                doc = Svg.SvgDocument.Open<Svg.SvgDocument>(imageStream);
+            }
+            catch
+            {
+                // Neither bitmap nor svg can load this, so that's an error.
+                throw new ArgumentException("Unable to decode image from supplied stream", nameof(imageStream));
```
ashahabov, line 155: Please use library custom class [SCException](https://github.com/ShapeCrawler/ShapeCrawler/blob/12982b7384a54285edf97fdded6820f6743245e4/src/ShapeCrawler/Exceptions/SCException.cs#L5) for exceptions.

ashahabov (changes requested):

ashahabov:
> Last two commits above were done to pass sonarcloud. While I agree with its feedback, and think they are good changes, this now leads to the new code being dissimilar from the existing code. Up to you which is more important: Keeping sonarcloud happy, or congruence with existing code.
> 
> I suppose another option would be to make these changes to the existing code as well, so now sonarcloud is happy, code is improved, AND old/new code matches nicely.


I suggest meeting the requirements of SonarCloud as much as possible, but it's better to avoid modifying code that isn't related to this task. This will help us prevent inflating the Pull Request.

jcoliz:
> > I suppose another option would be to make these changes to the existing code as well, so now sonarcloud is happy, code is improved, AND old/new code matches nicely.
> 
> I suggest meeting the requirements of SonarCloud as much as possible, but it's better to avoid modifying code that isn't related to this task. This will help us prevent inflating the Pull Request.

Right, making these changes to existing code would be a separate change, for sure, if you decided to go that route.

jcoliz:
@ashahabov I noticed this is still "In testing" in the project. I can say I've completed testing this with 200+ SVGs from many sources, and have reported all the issues found. In my own use, I've worked around every issue, and the results are stellar. So, I'd say this is looking good. We are going live on Tuesday with the results 😁

ashahabov:
@jcoliz, I believe, but I'd still like to try this out manually since this topic is new to me.

Reviewing PRs has a higher priority on my To-Do list🙂, so there is no time for another activities yet.

As a reminder, you can always if you need NuGet Package from the latest `master` you can get it from `https://www.myget.org/F/shape/api/v3/index.json` [Prerelease Version](https://github.com/ShapeCrawler/ShapeCrawler?tab=readme-ov-file#prerelease-version)

jcoliz:
@ashahabov No worries, just wanted to let you know how things were looking on my end.

I think I'm done with SVG PRs now , all known issues on my list addressed. Possibly will go after adding notes, but then you'll probably have a break from my PRs anyway.

Issue #677, opened 2024-05-12
Lots of compiler warnings


jcoliz:
@ashahabov Compiling generates many warnings. Would you take a PR to clean these up?

```
Logger\SCLogger.cs(23,9): warning CS0162: Unreachable code detected
Presentations\Slides.cs(150,30): warning CS0168: The variable 'sourcePresDoc' is declared but never used
Texts\TextParagraphPortion.cs(81,28): warning CS8629: Nullable value type may be null.
SlideMasters\ISlideMasterCollection.cs(32,13): warning CS0219: The variable 'number' is assigned but its value is never used
SlideMasters\SlideLayouts.cs(22,13): warning CS0219: The variable 'number' is assigned but its value is never used 
ShapeCrawler\Texts\Field.cs(86,28): warning CS8629: Nullable value type may be null. 
Charts\Chart.cs(20,32): warning CS0169: The field 'Chart.chartType' is never used
Colors\Color.cs(95,22): warning CS1572: XML comment has a param tag for 'result', but there is no parameter by that name
IShape.cs(129,96): warning CS1574: XML comment has cref attribute 'Audio' that could not be resolved 
ShapeCollection\IShape.cs(129,144): warning CS1574: XML comment has cref attribute 'Video' that could not be resolved 
```

ashahabov:
Yes, it would be good👍 But it is very desirable to do it through setting `.severity = error` in [.editorconfig](https://github.com/ShapeCrawler/ShapeCrawler/blob/master/src/ShapeCrawler/.editorconfig) _if possible_, otherwise as practice shows such a thing always falls into the `master`🙂

yogendrasgautam:
@jcoliz Is this issue still open, Can I start working on it ?

ashahabov:
@yogendrasgautam, thank you for the interest. 

Yes, the issue is open. There are still warnings left.

jcoliz:
@ashahabov @yogendrasgautam Fascinating! Which warnings? It's clean for me...

```
PS C:\Source\jcoliz\ShapeCrawler> dotnet build
  Determining projects to restore...
  All projects are up-to-date for restore.
  ShapeCrawler.Tests.Shared -> C:\Source\jcoliz\ShapeCrawler\test\ShapeCrawler.Tests.Shared\bin\Debug\netstandard2.0\Sh
  apeCrawler.Tests.Shared.dll
[...]

PR #680, opened 2024-05-12
Refactor many compiler warnings to errors

jcoliz:
Set severity to error for all warnings which were getting hit in current build. Attempted to fix them in least-intrusive-possible manner.

For #677

---

## PR-Codex overview
The focus of this PR is to remove unnecessary classes, update color handling in text portions, and adjust severity levels in the editor configuration file.

### Detailed summary
- Removed unnecessary logger classes
- Updated color handling in text portions
- Adjusted severity levels in the editor configuration file


src/ShapeCrawler/Logger/SCLogger.cs
```diff
@@ -20,6 +18,7 @@ public static void Send()
     {
         return;
 
+#if false
```
ashahabov, line 21: Could you please remove the `SCLogger` class and related code. In fact, it is not used.

src/ShapeCrawler/Texts/Field.cs
```diff
@@ -81,6 +82,11 @@ private Color ParseTextHighlight()
 
     private void UpdateTextHighlight(Color? color)
     {
+        if (color == null)
```
ashahabov, line 85: I think, it doesn't look logical: the method accepts nullable parameter but checks it for `null`. If we are going to do such a check, it is probably better to do it at a higher-level public method. It keeps the internal logic clean and reduces redundancy.

ashahabov (changes requested):

src/ShapeCrawler/Logger/SCLogger.cs
jcoliz: OK, sounds good will do. FYI, busy at work all week, so this will happen on weekend.

src/ShapeCrawler/Texts/Field.cs
jcoliz: Yup. In fact, probably would be better to push non-nullable all the way up to the public interface right? Null has no useful meaning here. In the original code, sending in null will just throw a exception on the cast to (non-nulllable) `Color`

ashahabov: Right, but for me, it is not yet clear what is better way in Setter of nullable _property_. Maybe something like the below code.

image

Feel free to make suggestions for more elegant refactoring.

jcoliz: How about making `TextHighlightColor` just be non-nullable? In fact, `ParseTextHighlight()` already returns a non-nullable `Color`. Setter is going to throw an exception. Nullable `Color` seems incorrect in this case.

ashahabov: Non-nullable `TextHighlightColor` also suits, but then I would suggest introducing some new `Color.NoColor`:

image

I believe it should be possible for the user to know if the text is highlighted.

jcoliz: > I believe it should be possible for the user to know if the text is highlighted.

Totally agree. But that would be outside the scope of a warning-fix PR, yes?

Furthermore, from looking at the code, I believe it's possible for user to know the text is highlighted or not, because TextHighlightColor will return Color.Transparent if there is no text highlight element. There doesn't seem to be a way to REMOVE the text highlight element, although perhaps setting it to Color.Transparent will have the same effect. This seems like a good place for someone to dig in who is interested in text highlighting.

Also, I am seeing now perhaps why Color was nullable to start with. `ParagraphLineBreak` implements `IParagraphPortion`, but cannot provide `Hyperlink` or 'TextHighlightColor`. So returns null, and throws on set. I changed to throw on get as well.

ashahabov: You are right: it is outside the scope of this PR. Let's let `Color.Transparent` act as “No Color” for the meantime.

There is an idea to name properties, actions, etc. as they are represented in PowerPoint. We will research this in the separate issue https://github.com/ShapeCrawler/ShapeCrawler/issues/690

Issue #686, opened 2024-05-16
Sub-pixel precision for shape position and size


jcoliz:
I'm doing a lot of work that revolves around precisely centering groups of shapes on both axes. Integer precision is insufficient for this. You can really tell looking at the slides that the centering is off by one pixel here and there. 

This has been fixed perfectly by using sub-pixel positioning and sizing. Prototype here: https://github.com/jcoliz/ShapeCrawler/tree/topic/decimal-positioning . I did this using decimals, but afterward I'm not convinced decimal is any better than double in this application.

Would like to contribute this back. Question for @ashahabov , how would you prefer it done?
1. Change int to double universally for size and position
2. Add separate properties for backward compat (this is how I did it my prototype)
3. Something else?

ashahabov:
Let's convert `IPosition.X` and other position and size properties into `decimal` type. In the future, if there are any complaints about performance, we will add some `IPosition.FloatX`.

PR #688, opened 2024-05-16
Convert position and size properties into decimal type

jcoliz:
As discussed in #686. This is a wide-reaching change.

Of note, there were a handful of unit tests which required fairly large (+/- 5) approximation margins, which seems odd and worthy of further study.

This did work fine for my fairly limited use cases.

---

## PR-Codex overview
This PR updates various properties and methods across the codebase to use `decimal` instead of `int` or `double` for more precise calculations and measurements.

### Detailed summary
- Changed properties `X`, `Y`, `Width`, `Height`, `LeftMargin`, `RightMargin`, `TopMargin`, `BottomMargin` to `decimal`
- Updated methods to use `decimal` parameters for more accurate calculations

> The following files were skipped due to too many changes: `src/ShapeCrawler/ShapeCollection/ShapeSize.cs`, `src/ShapeCrawler/ShapeCollection/RootShape.cs`, `test/ShapeCrawler.Tests.Unit/SlideMasterTests.cs`, `src/ShapeCrawler/ShapeCollection/SlideShapeOutline.cs`, `test/ShapeCrawler.Tests.Unit/TextFrameTests.cs`, `src/ShapeCrawler/Shared/UnitConverter.cs`, `test/ShapeCrawler.Tests.Unit/ShapeTests.cs`, `src/ShapeCrawler/Texts/TextFrame.cs`

### DIFF
--- a/src/ShapeCrawler/Drawing/IShapeOutline.cs
+++ b/src/ShapeCrawler/Drawing/IShapeOutline.cs
@@ -11,7 +11,7 @@ public interface IShapeOutline
     /// <summary>
     ///     Gets or sets outline weight in points.
     /// </summary>
-    double Weight { get; set; }
+    decimal Weight { get; set; }
 
     /// <summary>
     ///     Gets or sets color in hexadecimal format. Returns <see langword="null"/> if outline is not filled.
--- a/src/ShapeCrawler/Positions/IPosition.cs
+++ b/src/ShapeCrawler/Positions/IPosition.cs
@@ -9,10 +9,10 @@ public interface IPosition
     /// <summary>
     ///     Gets or sets x-coordinate of the upper-left corner of the shape.
     /// </summary>
-    int X { get; set; }
+    decimal X { get; set; }
 
     /// <summary>
     ///     Gets or sets y-coordinate of the upper-left corner of the shape in pixels.
     /// </summary>
-    int Y { get; set; }
+    decimal Y { get; set; }
 }
--- a/src/ShapeCrawler/Positions/Point.cs
+++ b/src/ShapeCrawler/Positions/Point.cs
@@ -11,7 +11,13 @@ public readonly ref struct Point
         this.X = x;
         this.Y = y;
     }
-    
+
+    internal Point(decimal x, decimal y)
+    {
+        this.X = (int)x;
+        this.Y = (int)y;
+    }
+
     /// <summary>
     ///     Gets the X coordinate.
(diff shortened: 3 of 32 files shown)


ashahabov:
> Of note, there were a handful of unit tests which required fairly large (+/- 5) approximation margins, which seems odd and worthy of further study.

We have all the previous calculations in `int`, after all. We didn't even round the numbers, and `int` type just roughly trimmed them. Or do you think that even with this, the new values are too far different?

I would consider continuing to use `.Should().Be()` in test assertions with the new values instead of `.Should().BeApproximately()`.

jcoliz:
> > Of note, there were a handful of unit tests which required fairly large (+/- 5) approximation margins, which seems odd and worthy of further study.
> 
> We have all the previous calculations in `int`, after all. We didn't even round the numbers, and `int` type just roughly trimmed them. Or do you think that even with this, the new values are too far different?
> 
> I would consider continuing to use `.Should().Be()` in test assertions with the new values instead of `.Should().BeApproximately()`.

Right, for those very-large deltas, I need to go back and look through each one specifically to see whether the new code caused a problem, or the new code is RIGHT and the test needs to be updated. This is the "further study" mentioned there. It seems so different, that I suspect an actual underlying problem. Adding the very large margin on those was just to get the tests to pass, and be able to submit so I could see what SonarCloud was going to complain about.

In most cases, the `.Should().BeApproximately()` margin is 1m, so just is rounding. Would you still want to specify the precise `.Should().Be()` value for those, out to the very very many digits of precision we'll have now? My concern is that this may make cause future test failures when use wouldn't notice difference.

ashahabov:
Then maybe it's better to go back to your original suggestion: adding a new properties `DecimalX`, `DecimalWidth` ... ?🙂

jcoliz:
Ok, so I going back through the large deltas. It turns out that moving to decimal sizes universally is a big improvement in correctness in these cases. In the old code, the top/bottom margins are set to "4" pixels, in new code, it's really "4.913". Then we compound that by using the top/bottom margin SIX times, resulting in a 6 * 0.913 pixel error, or 5.4798 combined error.

jcoliz:
Ok, I updated the size/position tests to use approximation ranges of 0.01, unless the SUT returned whole numbers, then used `Should().Be()`. If you **really** want `Should().Be(172.83622047244094488188976378M)`, please let me know and I can go back to update them.

Also I found many more places where decimal size/positions could be used.

PR #691, opened 2024-05-19
When adding SVG picture, handle case where dimensions are SvgUnitType.Percentage 

### LINKED ISSUE #687: SVG AddPicture: Non-square logos without width/height attributes are sized as squares
1. Create an SVG image which is wider than it is tall
2. Ensure it has no `width` or `height` attributes on the `svg` tag
3. Add it to a new slide, using ShapeCrawler
4. Load the resulting slide into PowerPoint
5. Notice that the shape is square, so the image is squished horizontally
6. Now with PowerPoint open, import that SVG manually
7. Notice that the resulting size matches the `viewbox` attribute, so PowerPoint inferred the size from that

More on this: If the SVG doesn't even have a `viewbox` attribute, PowerPoint will use 500x500 pixels. That's probably not what user wants, but at least user can SEE something.

jcoliz:
This fixes #687 as-written. I still need to test what happens when SVG has also no `viewbox`.

---

## PR-Codex overview
This PR updates SVG handling in ShapeCrawler. It adds support for SVG images without explicit dimensions and optimizes size calculations.

### Detailed summary
- Added support for SVG images without width and height tags
- Improved SVG size calculation for rasterization
- Introduced methods to convert inch to pixel and point to pixel
- Updated tests for SVG image handling


src/ShapeCrawler/ShapeCollection/SlideShapes.cs
```diff
@@ -582,8 +617,10 @@ private void AddPictureSvg(SvgDocument image)
         // Fix up the sizes
         var xEmu = UnitConverter.HorizontalPixelToEmu(100);
         var yEmu = UnitConverter.VerticalPixelToEmu(100);
-        var cxEmu = UnitConverter.HorizontalPixelToEmu(width);
-        var cyEmu = UnitConverter.VerticalPixelToEmu(height);
+
+        // TODO: Refactor this all to decimal when that PR lands
```
ashahabov, line 621: If I'm not mistaken, this is about “X” PR https://github.com/ShapeCrawler/ShapeCrawler/pull/688 ?

jcoliz: Right. When `HorizontalPixelToEmu()` changes over to decimal, this will need a little cleanup of casts here.

jcoliz: This is all updated now

test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
```diff
@@ -445,6 +445,52 @@ public void AddPicture_too_large_adds_picture()
...
+        var pres = new Presentation();
+        var shapes = pres.Slides[0].Shapes;
+        var image = TestHelper.GetStream("test-vector-image-wide.svg");
+        image.Position = 0;
+
+        // Act
+        shapes.AddPicture(image);
+
+        // Assert
+        shapes.Should().HaveCount(1);
```
ashahabov, line 461: Please remove this assertion. I believe it has little use in this test.

It is generally recommended to use as less assertion statements as possible in a unit test. Thus, for example, in this test, we focus more on the fact that this test is about the size of the added picture.

ashahabov (changes requested):

test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
jcoliz: is done

PR #694, opened 2024-05-20
Fix problems with text in SVG pictures

### LINKED ISSUE #681: SVG AddPicture: Incorrectly renders text in SVG
1. Create an SVG with text
2. Add that picture to a new presentation using ShapeCrawler
3. Load the result into PowerPoint
4. Notice that the beginning of the text is cut off

1x1
1x1-svg-wrong

```c#
using ShapeCrawler;

var pres = new Presentation();
var slide = pres.Slides[0];
var shapes = slide.Shapes;

var filename = "1x1.svg";
using var stream = File.OpenRead(filename);
shapes.AddPicture(stream);

pres.SaveAs($"out/{filename}.pptx");
```

jcoliz:
Fixes #681

As noted in the issue, the root cause was that `AddPictureSvg()` re-writes the SVG stream. This introduced a problem in `<text>` tags. The solution is to simply save in PowerPoint the SVG data we were originally given.

Also note that I was unable to craft an automated test for this. I did add a test which includes a manual verification step, and left it as `Explicit`. If you want to hear the long list of ideas I tried, I will gladly share!

Finally, I will mention that this may have some conflict with #691, which I can resolve when that lands.

---

## PR-Codex overview
The PR focuses on adding support for SVG images with text in PowerPoint presentations. 

### Detailed summary
- Added support for loading SVG images with text in PowerPoint presentations.
- Updated unit tests to validate SVG images with text.
- Modified `AddPictureSvg` method to accept a `Stream` parameter for SVG images.


jcoliz:
SonarCloud doesn't like me leaving an unresolved TODO. 

```c#
        // TODO: Consider higher resolution. PowerPoint seems to rasterize SVGs at 384-dpi, so 4x
        // our dpi.
```

This could be moved to a (very low priority) issue if you prefer.

ashahabov:
> SonarCloud doesn't like me leaving an unresolved TODO.
> 
> ```cs
>         // TODO: Consider higher resolution. PowerPoint seems to rasterize SVGs at 384-dpi, so 4x
>         // our dpi.
> ```
> 
> This could be moved to a (very low priority) issue if you prefer.

I have to admit that I agree with SonarCloud. Unfortunately, very often TODO items live forever in code if they are not turned into an issue with at least some minimal description.

test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
```diff
@@ -415,6 +415,32 @@ public void AddPicture_too_large_adds_svg_picture()
...
+
+        // Act
+        shapes.AddPicture(image);
+        pres.Validate();
+        var tempdir = Environment.GetEnvironmentVariable("TEMP") ?? throw new ApplicationException("TEMP directory not found");
+        pres.SaveAs($"{tempdir}\\AddPicture_svg_with_text_matches_reference.pptx");
+
+        // Assert
+        // This test must be checked manually. Load up the pptx saved above, and visually inspect
+        // the images to ensure they are identical.
```
ashahabov, line 441: I familiar with this situation. Ideally, of course, it would be to convert the shape with SVG-content to an image and compare this _actual_ image to the _expected_ image in some way.

I need some time to come up with at least some first step. I will back...

jcoliz: Indeed. So when I tried this with the tools at hand, the Svg library, it happily renders the SVG-content perfectly. Thus there is a subtle difference in how PowerPoint client renders svg versus how Svg library does so. I'm hesitant to accuse PowerPoint of having a bug, but certainly I'd posit Svg's way is BETTER.

A much hackier way would be to examine the SVG-content of the inserted image and verify that there is no whitespace between `<text>` and `<tbody>`. At least that's something.

jcoliz:
> > SonarCloud doesn't like me leaving an unresolved TODO.
> > ```cs
> >         // TODO: Consider higher resolution. PowerPoint seems to rasterize SVGs at 384-dpi, so 4x
> >         // our dpi.
> > ```
> > This could be moved to a (very low priority) issue if you prefer.
> 
> I have to admit that I agree with SonarCloud. Unfortunately, very often TODO items live forever in code if they are not turned into an issue with at least some minimal description.

Ok, is done.

test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
ashahabov: The only way to get the result of the PowerPoint client render in a unit test is probably to use `Interop.PowerPoint` (COM). I've tried this in unit tests in production, and I have to say, it's horrible 😒. `Interop.PowerPoint` is really not suited for automation in backend code.

For now, that leaves your suggestion: do a comparison based on internal content. At my main job, we do similar things when checking the result of docx to HTML conversion. These tests are certainly far from ideal; Resilience To Refactoring is very low since they depend on implementation details.

You may try such a test. I hope there will be few such tests in the future.

jcoliz: Ok, so done. See what you think!

test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
```diff
@@ -415,6 +416,47 @@ public void AddPicture_too_large_adds_svg_picture()
         pres.Validate();
     }
 
+    [Test]
+    [Explicit]
+    public void AddPicture_svg_with_text_matches_reference()
```
ashahabov, line 421: I've run locally this test in the `master` branch: it passes, but must fall.

ashahabov, line 421: What about such edits?

```C#
[Test]
public void AddPicture_svg_with_text_matches_reference()
{
    // ARRANGE

    // This presentation contains the same SVG we're adding below, manually
    // dragged in while running PowerPoint
    var pres = new Presentation(StreamOf("055_svg_with_text.pptx"));
    var shapes = pres.Slides[0].Shapes;
    var image = TestHelper.GetStream("1x1.svg");
    image.Position = 0;

    // ACT
    shapes.AddPicture(image);

    // ASSERT
    var picture = (IPicture)shapes.First(shape => shape.Name.StartsWith("Picture"));
    var xml = new XmlDocument { PreserveWhitespace = true };
    xml.LoadXml(picture.SvgContent);
    var textTagRandomChild = xml.GetElementsByTagName("text").OfType<XmlElement>().First().ChildNodes.Item(0);
    textTagRandomChild.Should().NotBeOfType<XmlSignificantWhitespace>("Text tags must not contain whitespace");
    
    // The above assertion does guard against the root cause of the bug 
    // which led to this test. However, the true test comes from loading
    // these up in PowerPoint and ensure the added image looks like the
    // existing image.
    pres.Validate();
}
```

Regard `xml.GetElementsByTagName("text").OfType<XmlElement>().First().ChildNodes.Item(0)`,

I think using statements as `for`, `if`, etc. in unit test make the test much more difficult to understand. If there is a case when, for example, `.ChildNodes.Item(0).Should().NotBeOfType<XmlSignificantWhitespace>()` is `true`, but `.ChildNodes.Item(1).Should().NotBeOfType<XmlSignificantWhitespace>()` is `false`, we'd better let such an error appear in production😁 and we'd better write a separate test for such case with a better understanding of what's going on.

jcoliz: If this is how you'd prefer to do it, that's fine. The facts here are that text elements should not have ANY children which are whitespace. In this particular case, when it was broken, items 0 and 2 were whitespace. If something broke in the future where just item 1 was whitespace, that still would be broken.

PR #698, opened 2024-05-28
Change font sizes to decimal type

### LINKED ISSUE #653: Change Font to double, not Int
In Powerpoint it is possible to set the font size as 8.3 or 7.2 or 10.8

However, Ifont.Size (used for a table like this: table[i, j].TextFrame.Paragraphs.First().Portions.First().Font.Size) is defined as an integer.

Thanks,
Serge

jcoliz:
Using decimal type to match recent changes to other size values.

Note that existing font size getter tests already had static test assets with fractional point text, so just needed to update tests to account for those. Modified setter tests to set decimal values.

Fixes #653

---

## PR-Codex overview
This PR updates font size properties across multiple classes from `int` to `decimal` for more precise calculations and consistency.

### Detailed summary
- Updated font size properties from `int` to `decimal` for precision and consistency.
- Updated font size calculations and updates to use `decimal` data type.
- Updated tests to reflect changes in font size data type.
- Adjusted font size calculations for better accuracy in various classes.

### DIFF
--- a/test/ShapeCrawler.Tests.Unit/FontTests.cs
+++ b/test/ShapeCrawler.Tests.Unit/FontTests.cs
@@ -347,7 +347,7 @@ public class FontTests : SCTest
         var font = pres.Slides[slideNumber - 1].Shapes.GetByName(shapeName).TextFrame!.Paragraphs[0].Portions[0].Font;
         var mStream = new MemoryStream();
         var oldSize = font.Size;
-        var newSize = oldSize + 2;
+        var newSize = oldSize + 2.4m;
 
         // Act
         font.Size = newSize;
--- a/test/ShapeCrawler.Tests.Unit/FontTests.cs
+++ b/test/ShapeCrawler.Tests.Unit/FontTests.cs
@@ -295,20 +295,20 @@ public class FontTests : SCTest
     [Test]
     [MasterShape("001.pptx", "Freeform: Shape 7", 18)]
     [SlideShape("020.pptx", 1, 3, 18)]
-    [SlideShape("015.pptx", 2, 61, 18)]
+    [SlideShape("015.pptx", 2, 61, 18.67)]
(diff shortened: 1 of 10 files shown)


jcoliz:
SonarCloud doesn't like the complexity of `PortionFontSize.IFontSize.Size()`. While I agree, please note that this issue was not introduced in this change.

test/ShapeCrawler.Tests.Unit/FontTests.cs
```diff
@@ -295,20 +295,20 @@ public void LatinName_Setter_sets_font_for_the_latin_characters(IShape shape)
...
+    [SlideShape("014.pptx", 2, 5, 21.77)]
     [SlideShape("012_title-placeholder.pptx", 1, "Title 1", 20)]
-    [SlideShape("010.pptx", 1, 2, 15)]
+    [SlideShape("010.pptx", 1, 2, 15.39)]
     [SlideShape("014.pptx", 4, 5, 12)]
     [SlideShape("014.pptx", 5, 4, 12)]
     [SlideShape("014.pptx", 6, 52, 27)]
     [SlideShape("autoshape-case016.pptx", 1, "Text Placeholder 1", 28)]
-    public void Size_Getter_returns_font_size(IShape shape, int expectedSize)
+    public void Size_Getter_returns_font_size(IShape shape, double expectedSize)
```
ashahabov, line 311: What about converting `expectedSize` into `decimal` type?

jcoliz: Sadly, this doesn't seem possible. C# does not allow `decimal` as an [attribute parameter type](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/attributes#2224-attribute-parameter-types). NUnit doesn't seem to want to convert it in this case, either. Happy to hear if you have a solution for this!

PR #704, opened 2024-06-01
Handle no-fill shape outlines

### LINKED ISSUE #703: Setting shape outline color to `null` causes presentation validation to fail
## Repro Steps

This test:
```c#
    [Test]
    [TestCase("autoshape-grouping.pptx", 1, "TextBox 6")]
    public void Color_Setter_null_removes_outline_color(string file, int slideNumber, string shapeName)
    {
        // Arrange
        var pres = new Presentation(StreamOf(file));
        var shape = pres.Slides[slideNumber - 1].Shapes.GetByName(shapeName);
        var outline = shape.Outline;
        
        // Act
        outline.HexColor = null;

        // Assert
        outline.HexColor.Should().BeNull();
        pres.Validate();
    }
```

## Result

Causes this validation failure:

```
"The attribute 'val' has invalid value ''. The attribute value cannot be empty."
```

## Expected result

Should remove the outline, as promised by IShapeOutline

```c#
    /// <summary>
    ///     Gets or sets color in hexadecimal format. Returns <see langword="null"/> if outline is not filled.
    /// </summary>
    string? HexColor { get; set; }
```

jcoliz:
This PR fixes 2 related issues.

First, shapes with no outline (including newly-created shapes) were erroneously reported as having outlines of color "000000", when they definitely were not solid black.

Second, attempting to set outline color to the *valid* value of `null`, led to presentation validation error. This is documented in the below-mentioned issue.

To fix the second part, I also implemented ability to set existing outline to no-fill.

Fixes #703

---

## PR-Codex overview
The focus of this PR is to enhance shape outline manipulation in ShapeCrawler by adding methods to set hex color and remove outline.

### Detailed summary
- Added `SetHexColor` and `SetNoOutline` methods to `IShapeOutline` interface.
- Implemented `SetHexColor` and `SetNoOutline` methods in `SlideShapeOutline` class.
- Updated tests to reflect changes in outline color manipulation.

### DIFF
--- a/src/ShapeCrawler/Drawing/IShapeOutline.cs
+++ b/src/ShapeCrawler/Drawing/IShapeOutline.cs
@@ -14,7 +14,17 @@ public interface IShapeOutline
     decimal Weight { get; set; }
 
     /// <summary>
-    ///     Gets or sets color in hexadecimal format. Returns <see langword="null"/> if outline is not filled.
+    ///     Gets color in 6-digit hexadecimal format. Returns <see langword="null"/> if outline is not filled.
     /// </summary>
-    string? HexColor { get; set; }
+    string? HexColor { get; }
+
+    /// <summary>
+    ///     Sets color in 6-digit hexadecimal format.
+    /// </summary>
+    void SetHexColor(string value);
+
+    /// <summary>
+    ///     Sets shape outline to "No outline".
+    /// </summary>
+    void SetNoOutline();
 }
--- a/src/ShapeCrawler/ShapeCollection/SlideShapeOutline.cs
+++ b/src/ShapeCrawler/ShapeCollection/SlideShapeOutline.cs
@@ -75,7 +75,22 @@ internal sealed class SlideShapeOutline : IShapeOutline
 
     private void Remove()
     {
-        this.sdkTypedOpenXmlCompositeElement.RemoveAllChildren<A.Outline>();
+        // Removing an outline really means ensuring that the shape has a 
+        // 'NoFill' outline. If we just REMOVE the outline, then the style
+        // will take over and give it the default outline, which is not
+        // what we want
+        A.Outline? outline = this.sdkTypedOpenXmlCompositeElement.GetFirstChild<A.Outline>();
+        if (outline is null)
+        {
+            outline = new A.Outline();
+            this.sdkTypedOpenXmlCompositeElement.AppendChild(outline);
+        }
+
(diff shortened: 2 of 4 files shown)


jcoliz:
Another approach, which is maybe a bit more akin to experience in PowerPoint client, is to make an `SetNoOutline()`. Still return `null` from color getter if no solid fill set. Throw an explicit exception on color setter, pointing user to `SetNoOutline()` if they want to remove the outline.

ashahabov:
> Another approach, which is maybe a bit more akin to experience in PowerPoint client, is to make an `SetNoOutline()`. Still return `null` from color getter if no solid fill set. Throw an explicit exception in color setter for `null` values, pointing user to `SetNoOutline()` if they want to remove the outline.

That was my first thought, too.

Maybe it's not relevant to this PR, but I think a redesign of something like the following for `IShapeOutline` would be beneficial:

```C#
string? HexColor { get; }
void SetHexColor(string hex);
void SetNoOutline();
```

Lately, I don't like the Setter in nullable property🙂

jcoliz:
> Maybe it's not relevant to this PR, but I think a redesign of something like the following for `IShapeOutline` would be beneficial

Yah, I think this is relevant here because it's all the area this PR is touching.

> Lately, I don't like the Setter in nullable property🙂

I agree in this case, if it's only going to throw on null setting then, it's not so good. In a broad case where `null` is a perfectly reasonable value, is there a more comprehensive argument against a Setter in nullable property?

ashahabov:
> In a broad case where `null` is a perfectly reasonable value, is there a more comprehensive argument against a Setter in nullable property?

I'm not sure if this is _another_ argument, but let me describe my understanding of nullable in this case: When the developer introduces a nullable property with a getter and setter, it says, "It can return `null`, and it can accept `null`". However, we throw an exception if the user sets it to `null`. This looks like we "lied", or at least it's not an _exceptional_ situation.

jcoliz:
> I'm not sure if this is _another_ argument, but let me describe my understanding of nullable in this case: When the developer introduces a nullable property with a getter and setter, it says, "It can return `null`, and it can accept `null`". However, we throw an exception if the user sets it to `null`. This looks like we "lied", or at least it's not an _exceptional_ situation.

Ok, yes, absolutely agree in this case. I was wondering if you were thinking that null-accepting setters were always bad in all cases, for some reason. For sure here, the interface lies by accepting a null setter, as you say.

Anyway, this has been updated. Please do take a look when you have a free moment.

Issue #798, opened 2024-12-13
Crop picture to shape


jcoliz:
Need to crop pictures to shapes. Specifically a top rounded rectangle

The following is found within a `p:pic` created by one of our designers:

```xml
<p:spPr>
    <a:xfrm>
        <a:off x="4185868" y="1372955"/>
        <a:ext cx="3815133" cy="387350"/>
    </a:xfrm>
    <a:prstGeom prst="round2SameRect">
        <a:avLst>
            <a:gd name="adj1" fmla="val 22122"/>
            <a:gd name="adj2" fmla="val 0"/>
        </a:avLst>
    </a:prstGeom>
</p:spPr>
```

ashahabov:
A possible new API to crop picture:

```C#
var pres = new Presentation("pre.pptx");
IPicture picture = pres.Slide(1).Picture("Picture 1");

picture.Crop(left:0.1, right:0.1, top:0.2, bottom:0.2); // Left and Right Crop is 10%, Top and Bottom Crop is 20%
```

jcoliz:
The `picture.Crop` API noted above would allow for setting the crop frame, which is good. However, it's not enough for this issue. We need to be able to specify the geometry and corner size for this crop also.

ashahabov:
Hmm, I didn't know that PowerPoint has feature "Crop to Shape". I think that's what you're talking about:

![image](https://github.com/user-attachments/assets/40d8256b-324e-4f69-b265-eb3a13dcb173)

jcoliz:
Yup, that's the one. User can still set the cropping frame, as well as apply preset geometry. And if you choose a rounded rectangle, you'll notice there is a corner size handle as well.

jcoliz:
Updated: I now think the clipping frame comes from `a:srcRect` within the `p:blipFill`

```xml
<p:blipFill rotWithShape="1">
    <a:blip r:embed="rId3">
        <a:alphaModFix amt="20000"/>
    </a:blip>
    <a:srcRect l="59" t="-295" r="19636" b="81332"/>
    <a:stretch/>
</p:blipFill>
```

jcoliz:
Side note: As discussed in #804, the range of the cropping frame should be changed from 0.0-1.0 to 0-100 instead.

PR #801, opened 2024-12-13
Basic image cropping

jcoliz:
Allows user to get/set the rectangular cropping frame on a picture.

Note that this PR doesn't deal with preset geometry. Still, it handles the basic cropping use case, which is worthy to stand alone as a feature.

Part of #798. 

---

## PR-Codex overview
This PR introduces a new feature for cropping images in the `ShapeCrawler` library. It adds a `CroppingFrame` property to the `IPicture` interface and implements its functionality in the `Picture` class. Additionally, it includes unit tests to verify the cropping behavior.

### Detailed summary
- Added `CroppingFrame Crop { get; set; }` to the `IPicture` interface.
- Implemented `Crop` property in the `Picture` class with logic to get and set cropping values.
- Created a new `CroppingFrame` record to represent cropping dimensions.
- Added unit tests for `Crop` getter and setter in `PictureTests`.


src/ShapeCrawler/Drawing/CroppingFrame.cs
```diff
@@ -0,0 +1,77 @@
...
+        aSrcRect.Top = ToHundredThousandths(this.top);
+        aSrcRect.Bottom = ToHundredThousandths(this.bottom);        
+    }
+
+    /// <summary>
+    ///     Parse a string value into a cropping frame.
+    /// </summary>
+    /// <param name="input">All four frame values separated by commas.</param>
+    /// <returns>Parsed frame.</returns>
+    public static CroppingFrame Parse(string input)
```
ashahabov, line 39: I'm in favor of avoiding the use of public statics whenever possible, which is a bad thing if you're trying to write in OOP style. Such not complicated parsing can be left for the library's client code.

Then using the API as shown below, simple, I think:

```C#
[TestCase("059_crop-images.pptx", 1, "None", 0, 0, 0, 0)]
public void Crop_getter_gets_expected_values(
    string file, 
    int slideNumber, 
    string pictureName, 
    int expectedLeftCrop, 
    int expectedRightCrop, 
    int expectedTopCrop, 
    int expectedBottomCrop)
{
    // Arrange
    var pres = new Presentation(StreamOf(file));
    var picture = pres.Slide(slideNumber).Picture(pictureName);

    // Act
    var crop = picture.Crop;

    // Assert
    crop.Left.Should().Be(expectedLeftCrop);
    crop.Right.Should().Be(expectedRightCrop);
    crop.Top.Should().Be(expectedTopCrop);
    crop.Bottom.Should().Be(expectedBottomCrop);
}

[TestCase(0, 0, 0, 0)]
public void Crop_setter_sets_expected_values(int leftCrop, int rightCrop, int topCrop, int bottomCrop)
{
    // Arrange
    var pres = new Presentation(StreamOf("059_crop-images.pptx"));
    var picture = pres.Slide(1).Picture("None");

    // Act
    picture.Crop = new CroppingFrame(leftCrop, rightCrop, topCrop, bottomCrop);

    // Assert
    picture.Crop.Left.Should().Be(leftCrop);
    picture.Crop.Right.Should().Be(rightCrop);
    picture.Crop.Top.Should().Be(topCrop);
    picture.Crop.Bottom.Should().Be(bottomCrop);
}
```

What do you think?

jcoliz: If that's how you prefer it, then that's how we shall do it.

BTW, you did take the easy route with 0,0,0,0. What would you do with 0.1,0.2,0.3,0.4? TestCase can't send in decimals, so now either we have to send in ints and divide them or strings and convert them. Now, we have to repeat that pattern in both tests and any future tests. So I thought it was better to DRY this and put this conversion in just one place. Also I'm a fan of hiding messy details.

Second, this same pattern follows with `FromSourceRectangle`. What would you prefer there? Constructing out of components means we have to push the per-cent-mille (hundredthousanths) conversions up to the caller, which seems like needless exposure of inner workings.

If it's OOP strictness you're after, we could have three different constructors instead. One fully qualified, one with a string, and one with a source rectangle.

I'd also point out that C# uses static `Type.Parse(string)` routinely, so I'd expect it's a pattern that developers are accustomed to working with.

ashahabov: > I'd also point out that C# uses static Type.Parse(string) routinely, so I'd expect it's a pattern that developers are accustomed to working with.

I agree. Perhaps such long-established things should be preferred over the desire for strict compliance with OOP🤔

> If it's OOP strictness you're after, we could have three different constructors instead. One fully qualified, one with a string, and one with a source rectangle.

This sounds good, however, except constructor for the source rectangle. I think you mean `A.SourceRectangle`? We've done this in the code already for another feature and I think it's wrong and should also be removed as it violates the basic idea of creating an object model over Open XML SDK. I mean, the _public_ APIs should not deal with entities of Open XML SDK.

PR #807, opened 2024-12-15
Set shape geometry to any arbitrary preset

### LINKED ISSUE #806: Proposed Preset Geometry refactor
Much in the same way that a `Position`  looks after an `A.Offset`, and a `ShapeSize` looks after an `A.Extents`, I'd like to collect everything that deals with an `A.PresetGeometry` into a single class. The word `Geometry` probably best describes this, but that's already taken by geometry type. Perhaps `PresetGeometry` or `ShapeGeometry`.

This comes about while thinking about #798. This would make cropping to shape really simple for user.

```c#
shapes.AddPicture(stream);
var picture = (IPicture)shapes.Last();
picture.Geometry = Geometry.RoundRectangle;
```

Of course, we could go and implement the above in `Picture`, but why not allow user to change geometry on anything with an `A.PresetGeometry`?

Would also clean up adding the rounded rectangles. Instead of having dedicated xml, and a dedicated creating method for every different kind of preset shape, we can just...

```
shapes.AddRectangle(stream);
var shape = shapes.Last();
shape.Geometry = Geometry.RoundRectangle;
```

Also it would allow for cleanup of `CornerSize`, which is oddly living in `Shape` now, while `AutoShape` has the geometry type, both of which deal with `A.PresetGeometry`

jcoliz:
This change allows user to set the geometry of shapes to any shape in the `Geometry` enum.

Probably the most invasive PR I've proposed, so please have a close look! Overall I think it was good to move A.PresetGeometry work into a dedicated class. Tried to follow the model used by other classes.

Fixes #806. This is the refactor I proposed in that issue.
Precursor for #798. That feature should be trivial now.

---

## PR-Codex overview
This PR focuses on refactoring the shape addition methods in the `ShapeCrawler` library, replacing specific rectangle shape methods with a more generic `AddShape` method, and enhancing geometry handling for shapes.

### Detailed summary
- Replaced `AddRectangle`, `AddRoundedRectangle`, and `AddTopCornersRoundedRectangle` methods with `AddShape`.
- Updated tests to reflect the new method and geometry handling.
- Enhanced `GeometryType` and `CornerSize` properties in `IShapeGeometry`.
- Removed deprecated shape XML resources.
- Improved shape geometry management in `ShapeGeometry`.


ashahabov (commented):
I'm ok with this new concept. There are just some suggestions:

- replace `shapes.AddRectangle(50, 60, 100, 70)` with `shapes.Add(50, 60, 100, 70, geometry = Geometry.Rectangle)` or `shapes.AddShape(50, 60, 100, 70, geometry = Geometry.Rectangle)`
- Rename the names of geometries, at least to the first ones we already use in our tests: `Round2SameRectangle` -> `TopCornersRoundedRectangle`, `Round2DiagonalRectangle` -> `DiagonalCornersRoundedRectangle`

jcoliz:
> I'm ok with this new concept. There are just some suggestions:
> 
>     * replace `shapes.AddRectangle(50, 60, 100, 70)` with `shapes.Add(50, 60, 100, 70, geometry = Geometry.Rectangle)` or `shapes.AddShape(50, 60, 100, 70, geometry = Geometry.Rectangle)`
> 
>     * Rename the names of geometries, at least to the first ones we already use in our tests: `Round2SameRectangle` -> `TopCornersRoundedRectangle`, `Round2DiagonalRectangle` -> `DiagonalCornersRoundedRectangle`

@ashahabov As said, so done. Thanks for reviewing!

PR #816, opened 2024-12-17
Insert solid fill at the start of the parent element

### LINKED ISSUE #793: SlideMasters Font Color Issue
It doesn't update the font color of the shape located on Slide Master.

Use test [Update_updates_font_color_of_master_shape](https://github.com/ShapeCrawler/ShapeCrawler/blob/33279bd768853f6fb29126ee6246590e4a3c4400/test/ShapeCrawler.Tests.Unit/FontColorTests.cs#L120-L136) to reproduce this bug:

```C#
[Test]
[Ignore("Should be fix with issue https://github.com/ShapeCrawler/ShapeCrawler/issues/793")]
public void Update_updates_font_color_of_master_shape()
{
    // Arrange
    var pres = new Presentation(StreamOf("061_font-color.pptx"));
    var fontColor = pres.SlideMasters[0].Shapes.GetByName("TextBox 1").TextBox.Paragraphs[0].Portions[0].Font!.Color;
    
    // Act
    fontColor.Update("#007F00");
    
    // Assert
    pres.SaveAs("result.pptx");
    // Open result.pptx in PowerPoint and check the color of the TextBox 1. It should be green.
    
    // TODO: add assertion
}
```

---

### Initial Message

I have a PowerPoint template with two slide masters with the same shape names.  The text can be changed on both slide masters with the same shape names, but when attempting to update the same color on these shape names, only the first slide master's color is updated.

[SlideMasters_Test.zip](https://github.com/user-attachments/files/18050382/SlideMasters_Test.zip)

blockingHD:
Comparing between a powerpoint edited file and the generated file from the "Update_updates_font_color_of_master_shape" you can see that the <a:solidFill> tag is in a different position. 
image

Doing a direct edit and moving the tag to the top fixes the issue when opening. 

I have changed from an append (putting it at the bottom) to an insert at position 0.

Fixes #793

---

## PR-Codex overview
This PR focuses on improving the handling and validation of font colors in PowerPoint presentations, specifically ensuring the correct structure of `SolidFill` elements and updating tests accordingly.

### Detailed summary
- Changed the way `aSolidFill` is appended to `aRunProperties` by using `InsertAt`.
- Added a new method `ValidateASolidFill` to check for valid `SolidFill` structures in slides and slide masters.
- Updated tests in `FontColorTests.cs` to validate the font color update process and removed an ignored test.


blockingHD:
@ashahabov I am unsure how to best write the test here as the test suit uses the abstracted version of the sdk doc which loses the position of tags. Should there be such a test or a comment above the insert explaining why with the original test in place for reference. 

Open to opinions 🙂

jcoliz:
@blockingHD This fixes #793?

We'll see what Adam thinks about testing. My view is that it's worth writing a different kind of test that digs into the underling XML types. The only accurate user-facing test involves loading a creted presentation file into the UI, which we can't do. Having a test that element is in correct place would ensure that someone else doesn't come along and insert something new at position 0, thereby re-causing this problem.

Another idea I've been thinking about is add the check into the presentation.Validate() logic, then the test is simply set up the situation and call pres.Validate(). The idea here is that Validate could/should check for things that the UI would/should check.

ashahabov:
We've already had a similar bug when tables didn't work because of wrong tags order. As @jcoliz mentioned, I added validation [ValidateATableRows()](https://github.com/ShapeCrawler/ShapeCrawler/blob/9b8d1eb3bd846b1a98330d231fbb13c1d4279b62/src/ShapeCrawler/Presentations/PresentationCore.cs#L115-L149). You can do the same.

Regard Test Assertion, I would suggest:

```C#
// Assert
var validation = () => pres.Validate();
validation.Should().NotThrow();
```

---

And yes, ideally and most properly, of course, it would be loading a created presentation file into the PowerPoint UI, which we can't do.

ashahabov:
> @blockingHD This fixes #793?

I think yes.

blockingHD:
Sorry. Yes this fixes #793. Added to the description.

blockingHD:
I have pushed what I would expect to work. However, the PowerPoint used for the test doesn't pass the schema check. I'm not sure how it was generated, so I can't say why that's the case. I have submitted the fix by changing the OOXML directly but this could be another bug.

The Fix is to remove the a:ln tags in both run properties. Not sure if there is some other cause for this but it works.
image

ashahabov:
@blockingHD, I've tested these changes: it works for me. Can I merge it?

blockingHD:
I'm happy for this to go in.

Issue #803, opened 2024-12-14
Add a collection of short example programs to the project


jcoliz:
I propose a top-level project folder named (e.g.) 'examples'. Under this would be numerous very small example programs each of which demonstrate a single API, in the spirit of a unit test. Each would output a presentation which we could then load into PowerPoint to confirm the API worked visually.

The problem this would solve is: Today when creating a new ShapeCrawler feature, there are three test modalities needed:
1. Create a presentation in PowerPoint, then load it in ShapeCrawler to ensure the getter gets the expected value, and the setter can modify it. We have a unit test pattern for this.
2. Round trip through the constructor/getter/setter. Again, unit tests are great for this.
3. Create a presentation in ShapeCrawler then load it in PowerPoint to ensure it looks right. Can't do this in unit tests. 😒

To cover the last one, I usually create a branch in the consuming app. But it would be better to have an isolated test case. Then also YOU have to create your own app to test the feature when the PR comes. It would be better if the PR comes with an example, and you can just run that and load it into PowerPoint yourself to confirm.

This could also be added to the unit tests to at least build and run the examples as one more safety net. Further, you could attach the outputs to the CI pipeline, so it would be easy to inspect whether a problem had cropped up in an API.

Finally, this serves a documentation purpose as well to help users in their quest to understand how to use the library.

ashahabov:
@jcoliz, I am still thinking about the suggestion. In the meantime, here are some thoughts:

**Folder `examples`.** We already had an `examples` folder where I sometimes wrote small tests. Later, I moved them to the [Wiki](https://github.com/ShapeCrawler/ShapeCrawler/wiki/Examples). _In my opinion_, it is easier for users to search for examples on the Wiki page than to look through folders. Actually, I am not happy with the current quantity and quality of the documentation. I am working on moving the documentation to [GitBook](https://shape.gitbook.io/shapecrawler). Any help or suggestions are welcome!

**Maintainer Testing.** I do manual testing when I review a Pull Request: I clone the contributor’s fork and check out its branch, or I do it after merging the Pull Request. In both situations, I call `IPresentation.SaveAs()` in the existing test to see the output in PowerPoint. But in a test like the one below, yes, I cannot do this. I must create a new temporary unit test for my manual testing. Maybe this is what you mean?


```C#
[Test]
[SlideShape("autoshape-grouping.pptx", 1, "TextBox 7", "SimSun")]
[...]

PR #826, opened 2024-12-23
Reuse existing image part

### LINKED ISSUE #809: Reuse existing image file
`AddPicture()` duplicates the image source.

Use test [AddPicture_should_not_duplicate_the_image_source_When_the_same_image_is_added_twice](https://github.com/ShapeCrawler/ShapeCrawler/blob/2f4cb91efe059a7ca695d2440339f1510f2ad1e8/test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs#L571-L588) to reproduce this bug:
```C#
[Test]
public void AddPicture_should_not_duplicate_the_image_source_When_the_same_image_is_added_twice()
{
    // Arrange
    var pres = new Presentation(StreamOf("008.pptx"));
    var shapes = pres.Slide(1).Shapes;
    var image = StreamOf("png image-1.png");
    
    // Act
    shapes.AddPicture(image);
    shapes.AddPicture(image);
    
    // Assert
    pres.SaveAs("output.pptx");
    // Check the folder output.pptx/ppt/media. This folder should contain only one image file.
    // TODO: Add assertion
}
```

---

### Initial Message

Doing some research into why my PPTX files are 30MB, and I discovered...

While running Powerpoint, if I add the same image into the document several times, the image is only added once to the file's `media` directory, and all shapes which use it refer to the same part ID in their blip.

However, using ShapeCrawler, every call to `AddPicture` adds the image _again_.

It would be great to have some mechanism where I could signify reuse. Or even better, if it happened automatically. So that the same image added twice would look up the existing part and refer to that instead of embedding a whole new one. I'd be happy to specify a string file name to use as a key.

jcoliz:
Re-use existing images instead of adding the same image bits to the presentation every time.

This PR adds a `MediaCollection` to track image parts by the hash of their data stream, then uses that to find existing matching image parts when adding images to slide.

Current state: Exhaustively tested. Ready to merge from my POV. Working through review comments.

Uses XML introspection to assert correct state, which goes against current conventions. However, I'd argue it's the best we have currently, and does allow for an automated test.

Fixes #809

---

## PR-Codex overview
This PR focuses on enhancing image handling in the `ShapeCrawler` library by introducing a `MediaCollection` class for image de-duplication and updating relevant methods to support this functionality.

### Detailed summary
- Added support for only PNG image mimetype in `IShapeFill`.
- Refactored `AddImagePart` method to return a tuple of `string` and `ImagePart`.
- Introduced `MediaCollection` class to manage image parts and prevent duplication.
- Updated constructors and methods in `Slide`, `SlideShapes`, and other classes to utilize `MediaCollection`.
- Added tests to ensure images are not duplicated when added multiple times or across different slides.


jcoliz:
More mature implementation. Handles dupes across slides now. Still doesn't handle dupes across sessions, that's next.

jcoliz:
Ok, that worked better than expected. 30MB file is down to 4.5MB.

Discovered a missing use case. The following code results in duplicated images in `media` folder:

```c#
ISlides.Add(ISlide)
```

jcoliz:
This is the line where the images are duplicated:

```c#
internal readonly ref struct WrappedPresentationPart
...
    internal void AddSlidePart(SlidePart slidePart)
...
        var addedSlidePart = this.presentationPart.AddPart(slidePart, rId);
```

It's not obvious how to circumvent this. Ergo, let's call this feature done, and consider a possible future refinement where images are de-duped when slides are copied.

test/ShapeCrawler.Tests.Unit/Helpers/SCTest.cs
```diff
@@ -62,6 +63,26 @@ protected static Presentation SaveAndOpenPresentation(IPresentation presentation
...
+    protected static PresentationDocument SaveAndOpenPresentationAsXml(IPresentation presentation)
+    {
+        var stream = new MemoryStream();
+        presentation.SaveAs(stream);
+        stream.Position = 0;
+
+        return PresentationDocument.Open(stream, true);
+    }
+
+    protected void SavePresentationFile(IPresentation presentation)
```
ashahabov, line 75: `SCTest.SaveResult()` is a duplication and can be removed?

test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
```diff
@@ -347,6 +348,69 @@ public void AddPicture_adds_svg_picture()
...
+
+        // Act
+        shapes.AddPicture(image);
+        shapes.AddPicture(image);
+
+        // Assert
+        var checkXml = SaveAndOpenPresentationAsXml(pres);
+        var imageParts = checkXml.PresentationPart.SlideParts.SelectMany(x=>x.ImageParts).ToArray();
+        imageParts.Length.Should().Be(2);
+        SavePresentationFile(pres);
```
ashahabov, line 367: As mentioned [here](https://github.com/ShapeCrawler/ShapeCrawler/issues/803#issuecomment-2564385180), please avoid using Hard Drive if the test doesn't explicitly test integration with Hard Drive.

test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
```diff
@@ -347,6 +348,69 @@ public void AddPicture_adds_svg_picture()
...
+        var image = StreamOf("test-vector-image-1.svg");
+
+        // Act
+        shapesSlide1.AddPicture(image);
+        shapesSlide2.AddPicture(image);
+
+        // Assert
+        var checkXml = SaveAndOpenPresentationAsXml(pres);
+        var imageParts = checkXml.PresentationPart.SlideParts.SelectMany(x=>x.ImageParts).Select(x=>x.Uri).ToHashSet();
+        imageParts.Count.Should().Be(2);
```
ashahabov, line 388: Is this assertion correct? Should be `imageParts.Count.Should().Be(1)`?

ashahabov (changes requested):

test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
jcoliz: SVG adds two parts. One for vector image and one for raster. I should add a reason to these assertions.

jcoliz: b6cbf1b09058a4efb42f27fed36df448d55f9b6c

test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
jcoliz: 47f0f07b52a7210f7317700ce7e35847010eedc0

test/ShapeCrawler.Tests.Unit/Helpers/SCTest.cs
jcoliz: No longer. I have removed the potential duplication in addressing below comment. 47f0f07b52a7210f7317700ce7e35847010eedc0

Issue #762, opened 2024-09-16
Unable to decode the image from the supplied stream


chris-askbrian:
I have quite weird problem as its pretty much really basic 

Im on ubuntu using net8.0 writing a classlib using ShapeCrawler v0.54.2

```
using ShapeCrawler;
namespace PptxLibrary
{
    public class PptxHelper
    {
        public void AddImgToPptx(string pptxPath, string imgPath)
        {
            var presentation = new Presentation(pptxPath);
            var slide = presentation.Slides[0];
            var imgStream = new FileStream(imgPath, FileMode.Open);
            slide.Shapes.AddPicture(imgStream);
            presentation.Save();
        }
    }
}
```
When using this code with an normal image it works without a problem, adding the picture and its done. With any kind of svg like this one: 
![svg](https://github.com/user-attachments/assets/5113f12f-d06f-4a5d-9703-098462a787c8)

I get a `ShapeCrawler.Exceptions.SCException: Unable to decode the image from the supplied stream.` Error

Not sure what I do wrong here (but tbh I'm not a skilled c# developer so I might have missed something but since It works with png I am not sure anymore)

ashahabov:
@chris-askbrian, I was unable to reproduce this error. Could you please attach this SVG file?

chris-askbrian:
[...]

PR #830, opened 2024-12-24
Changed to use Magik package to support Mac and Linux

blockingHD:
As an attempt to address the issue outlined in #823.

I changed the add picture method to use the [Magik.NET](https://github.com/dlemstra/Magick.NET?tab=Apache-2.0-1-ov-file) library. This supports all platforms with the AnyCPU package and is licenced under Apache 2.0. 

What are the thoughts on moving to this package? 
If this is acceptable, I will clean up the redundant code.

---

## PR-Codex overview
This PR focuses on enhancing the `ShapeCrawler` library by adding support for new image formats and improving the handling of SVG images. It updates dependencies, modifies build workflows, and includes tests to ensure proper functionality.

### Detailed summary
- Added a blank SVG file `test-vector-image-blank.svg`.
- Updated `.github/workflows/pull` to include `macos-14`.
- Added `Magick.NET-Q16-AnyCPU` and `Magick.NET.SystemDrawing` package references.
- Removed the `Svg` package reference.
- Modified `AddPicture` method in `SlideShapes.cs` to handle SVG images with `MagickImage`.
- Updated tests in `ShapeCollectionTests.cs` to verify image duplication logic.
- Added new tests for image handling and duplication scenarios.


jcoliz:
@blockingHD As long as new library does everything the old one does, with no user impact, it sounds like a good change.

As a rule, replacing dependent libraries should not change any tests. The fact that this PR changes test assertions is a concern.

One specific comment I have so far is that the PR uses the rasterized dimensions for the shape size. We should not do that. We should use the precise SVG size. Rasterizing will lose precision, rounding to whole pixels. Sub-pixel precision is necessary.

I have a library of hundreds of SVGs, which I can run through this when it's ready.

blockingHD:
> @blockingHD As long as new library does everything the old one does, with no user impact, it sounds like a good change.
> 
> As a rule, replacing dependent libraries should not change any tests. The fact that this PR changes test assertions is a concern.
> 
> One specific comment I have so far is that the PR uses the rasterized dimensions for the shape size. We should not do that. We should use the precise SVG size. Rasterizing will lose precision, rounding to whole pixels. Sub-pixel precision is necessary.
> 
> I have a library of hundreds of SVGs, which I can run through this when it's ready.

Good point around the test changes. I assumed that this library was creating the PNG tighter to the true size of the SVG, given that it wouldn't be a surprise if a border were added. However, this could equally be my code so I will take more of a look.

jcoliz:
Yeah that particular file is exactly 90.0 x 276.9623. In fact, I probably should have gotten rid of the approximation in the test. The extent comes from this element:

```xml
<circle
   r="45"
   cx="231.9623"
   cy="45"
   fill="#ff0000"
   id="circle1-6" />
```

Also, this PR changes the rasterization DPI to 384. My view is that one PR should change one thing, to aid in focused review. Also, that's a change I'd like to discuss a bit more. I'll put my comments in that issue.

blockingHD:
@jcoliz I have looked at this; the height and widths are uint's, so Magik is adding imprecision and in the wrong direction! 
image

What do we want to do?

P.S. I can split the PPI change into a separate PR. I will look after there is a call re-making this change.

jcoliz:
Ok, so I have another idea. What if we entirely removed SVG processing from ShapeCrawler full stop. This would put rasterization on the user. Not super ideal, but it would avoid these dependency issues. As a side note, I'm worried about taking a dependency on AnyCPU architecture. Despite the name, it's still an architecture, and could cause issues with Win Arm. Removing SVG processing would avoid that too.

This could be implemented pretty easily. Add a second AddPicture signature to accept a SVG parameter. It means user would have to rasterize SVGs themselves, and pass the rasterized version in as the `image`, adding the svg as well. For tests, we can just add the rasterized PNG's to the test project.

```c#
public void AddPicture(Stream image, Stream svg)
```

Likewise, if user cares about precise sizing, we can put it on them too. We can accept the whole-pixel-precision sizing that comes from the user's rasterized image as the default. 

I'm curious to know if anyone else but me is using SVGs in ShapeCrawler 😁. For me, I'm already loading and measuring the SVG, so rasterizing it is simple.

blockingHD:
I am with you there @jcoliz the more I am looking at libraries and working out the pros and cons of each, the more I feel maybe this library should be agnostic. It would be ideal to have a default implementation, but as this is an architecture-specific task, then, as you say, an AnyCPU implementation still has limitations.

Further to this, PowerPoint seems to be dealing with SVGs entirely differently from the implementation for SVGs with no dimensions. This might just be a Mac thing? Which in fairness gives more weight to this argument. 

Guessing we need @ashahabov to make a call?

jcoliz:
> PowerPoint seems to be dealing with SVGs entirely differently from the implementation for SVGs with no dimensions. 

Please elaborate. As part of writing SVG support, I did a ton of work on the various permutations of dimensionless SVGs. Each of the SVGs in my collection is created by someone different, so this leads to a lot of variability. My goal was to match PowerPoint's outcome, as per design goals of the library. So I'd love to know what differences you're seeing.

blockingHD:
When adding the dimensionless SVG it gets inserted like this in PowerPoint:
Screenshot 2024-12-24 at 22 29 23

With the generated image being 2000x2000:
Screenshot 2024-12-24 at 22 30 37

jcoliz:
Ah, yes. Ok I think I decided that was so broken and unusable, there's no way user would want that to happen. This is a case where I think the library works _better_ than the UI. No reason we can't just infer dimensions from contents.

Having experimented with this a bit now, I've noticed a few things.
* If dimensions are set as fractional units, ImageMagick does the right thing and rounds UP, so the rasterized image looks good.
* In this case with un-set dimensions, ImageMagick actually slices off the right-most pixel, which isn't great. However, does it really matter? Rasterized image seems to be a formality. IIRC PowerPoint re-rasterizes on load anyway.
* ImageMagick AnyCPU library is compiling to native Arm64 binary no problem on Windows, so, looks like this won't cause a problem for this platform.

Maybe this is all fine as-written. Once you think it's ready, I can try on a pile of real data.

jcoliz:
Unfortunately, it seems that the current released code is actually broken in this case. In the test `AddPicture_adds_svg_picture_no_dimensions`, if we add a `pres.SaveAs`, and then load that up in PowerPoint, it is completely wrong. The dimensions match the source image, but it's actually the 2000x2000 image you showed, shrunk down to the "correct" size, so way too much white space.

I think the right course of action is to declare adding a dimensionless SVG "unsupported, producing undefined results," and remove the test.

ashahabov:
> Guessing we need @ashahabov to make a call?

Hello @blockingHD, hello @jcoliz.

I apologize for the silence. I just got back from vacation with my family yesterday. I am currently checking this PR along with others... I will back with my feedback...

Thanks for doing this hard work with images!

blockingHD:
I think this is ready for testing with your collection of SVGs @jcoliz.

I believe the SVG validity check is correct in that you must have a width and height or a view box to determine the size of the image. This isn't possible in an XSD, I don't think, as it requires logic around attributes which isn't supported. Please let me know if I am wrong!

Other than that, the only other thought I have is to throw a custom exception for the SVG failure to make it easier to catch. Not sure what other people think on that one.

ashahabov:
I think both of you have more experience with SVG content than I do, so I’m ready to go with whatever you think is the best. I just have a few limitations that I’ll share at the end below. First, let me explain the problem as I understand it. Please correct me if I’m wrong:

- The used SVG library doesn’t work on ARM-based macOS. Switching to Magick.NET has been suggested to solve this issue.
- The rendering quality of Magick.NET is worse than the current SVG library.

---

We had a similar [issue](https://github.com/ShapeCrawler/ShapeCrawler/issues/762#issuecomment-2409086910) before where a developer had a problem on Ubuntu. We solved it also by running `sudo apt install -y libgdiplus`. I agree with @blockingHD that we should document such solutions.

I'm not sure regard suggested `AddPicture(Stream image, Stream svg)`. It would be good to keep it as `AddPicture(Stream image)`. Can’t we detect if the `Stream` is SVG content and handle the rasterization automatically?

With SkiaSharp, the SVG library, and now Magick.NET, we are using three dependencies for graphics. We should try to reduce this to two. Either keep the SVG library or use Magick.NET, but three dependencies seem too much.

jcoliz:
@ashahabov This PR should remove SVG library dependency.

It's an interesting exercise for the future to see if ImageMagick could also replace SK image lib.

Add picture API with two streams would only be if you wanted to entirely remove any dependency related to SVG processing and put it on the user. It's extra work for SVG user but less dependencies for user not using SVG.

Please give me a couple days to test. Will be back at work later this week and want to test more there.

ashahabov:
> It's an interesting exercise for the future to see if ImageMagick could also replace SK image lib.

Yeah, that would be perfect.

jcoliz (changes requested):
Couple comments from static review. Will try with images in next couple days.

jcoliz:
@blockingHD I don't think we can be thorough enough to test for all cases of missing height & width. I have an image that looks like this, which PowerPoint loads fine, but which triggers the "must have width or height" exception. IMO, we should accept undefined behavior here. Other solution is use a library, like SVG, which is likely to be sufficiently thorough, but obviously that's what we're trying to get away from.

```xml
<svg:svg
   version="1.1"
   id="Livello_1"
   x="0px"
   y="0px"
   viewBox="0 0 2744.5306 2553.2816"
   enable-background="new 0 0 2744.833 2553.333"
   xml:space="preserve"
   width="3.8889999in"
   height="3.618in"
```

src/ShapeCrawler/ShapeCrawler.csproj
```diff
@@ -69,6 +69,8 @@ This library provides a simplified object model on top of the Open XML SDK for m
   <ItemGroup>
     <PackageReference Include="DocumentFormat.OpenXml" Version="3.2.0" />
     <PackageReference Include="DocumentFormat.OpenXml.Linq" Version="3.1.1" />
+    <PackageReference Include="Magick.NET-Q16-AnyCPU" Version="14.2.0" />
+    <PackageReference Include="Magick.NET.SystemDrawing" Version="8.0.2" />
```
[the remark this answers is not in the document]
ashahabov: Yes.

src/ShapeCrawler/ShapeCollection/SlideShapes.cs
```diff
@@ -144,37 +146,49 @@ public void AddPicture(Stream image)
...
+            });
+            imageMagick.Format = MagickFormat.Png;
+
+            width = imageMagick.Width < 500 ? (int)imageMagick.Width : 500;
+            height = imageMagick.Height < 500 ? (int)imageMagick.Height : 500;
+
+            if (width > 500 || height > 500)
             {
-                throw new SCException("An error occurred while attempting to use GDI+ functionality. If you use Linux, you need to install a library that provides GDI+ support. You can do this with the following command: sudo apt install -y libgdiplus", ex);
+                imageMagick.Resize((uint)width, (uint)height);
```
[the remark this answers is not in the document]
blockingHD: The logic here was just wrong. Though the height and width were being set but the image still need to be resized. Corrected now, so we constrain it to 500px then resize if it is 500px. This is slightly inefficient because what if the image was 500px to begin with but hardly seems worth worrying about.

I did have it where the height and width were constrained in the if block but SonarQube complained about cognative complexity. Not sure this is better but the machine told me to do it 😅

ashahabov:
@blockingHD, let me remind that you can see the code style issues locally building Release configuration instead of Debug.

blockingHD:
> @blockingHD, let me remind that you can see the code style issues locally building Release configuration instead of Debug.

I have built using the release config and I am not getting this error. So a little confused.

ashahabov:
@blockingHD, you should remove these unused usings:

image

blockingHD:
@ashahabov Appologies, I was being a little silly. Did the merge in GitHub then never pulled the code back down. Should be fixed now :)

test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
```diff
@@ -526,26 +530,27 @@ public void AddPicture_adds_svg_picture_no_width_height_tags()
         picture.Width.Should().Be(280);
         pres.Validate();
     }
-
+    
+    //Test that the png image is created with transparent background
     [Test]
-    public void AddPicture_adds_svg_picture_no_dimensions()
```
ashahabov, line 531: Are we abandoning this test `AddPicture_adds_svg_picture_no_dimensions`?

@jcoliz, I've run this test `AddPicture_adds_svg_picture_no_dimensions` on the latest changes of actual PR. It fails: The actual values:

```C#
picture.Height == 90
picture.Width == 276
```

ashahabov:
@blockingHD, I've added testing on macOS 13 (later also Ubuntu) in workflow Pull Request. Could you please back merge `master` branch to see whether it breaks something.

test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
blockingHD: We had this failure earlier on. The new package doesn't calculate the size perfectly. However, due to how PowerPoint handles SVGs, the reality is that as soon as you save the PowerPoint, it gets recalculated incorrectly anyway. Hence, the suggestion of stating dimensionless SVGs produce undefined behaviour.

jcoliz: Yup. This test is not helpful. It's true, and it works, but the outcome is not what we want.

As an example, imagine a 200x100 solid black image. It will be correctly displayed as 200x100 in PowerPoint. However, the black area is scaled down to 20x5, and the remaining pixels are blank.

You can try by saving the result of this test to disk and then load in PowerPoint.

jcoliz:
FYI, ff7977199787c1b978263e786603869a5d605647 looks good to me. Tested against what I have, and all is well.

ashahabov:
@blockingHD, I've also finished testing, and I think it is ready to merge🎉 Please, back-merge `master` to resolve conflicts.

blockingHD:
@ashahabov This should be ready for approval and merging :)

.github/workflows/pull request.yml
```diff
@@ -10,28 +10,22 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ windows-latest, ubuntu-latest, macos-13 ]
```
ashahabov, line 13: Please preserve also `ubuntu-latest`.

test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
```diff
@@ -591,132 +573,7 @@ public void AddPicture_should_not_change_the_underlying_file_size()
         var addedPictureImage = shapes.Last<IPicture>().Image!;
         addedPictureImage.AsByteArray().Length.Should().BeLessThan(38000);
     }
-
-    [Test]
-    public void AddPicture_should_not_duplicate_the_image_source_When_the_same_png_image_is_added_twice()
```
ashahabov, line 596: Please, do not remove this and another below `AddPicture_***` tests

blockingHD: This looks to have happened in the merge conflict commit you did. Do you want me just to revert that?

ashahabov: Yes, try to make revert to keep these tests or maybe it would be easier to just copy past these tests from origin `master`.

.github/workflows/pull request.yml
```diff
@@ -10,28 +10,22 @@ jobs:
...
+        os: [ windows-latest, ubuntu-latest, macos-13, macos-14 ]
         framework_version: [ 'net9.0', 'net48' ]
         exclude:
-          - os: ubuntu-latest
-            framework_version: 'net48'
```
ashahabov, line 17: Revert also these changes to prevent trying to test on .NET Framework 4.8 on `ubuntu-latest`.

blockingHD: Should be sorted now.

test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
blockingHD: The diff is borderline unintelligible but I think this is sorted now?

ashahabov: I've pushed some changes in your branch `Add-SVG-Support-For-Mac-and-Linux`. Please pull.

Issue #842, opened 2025-01-02
Fix for #828 fails to load in PowerPoint


jcoliz:
Repro Steps

1. Edit this test: `LatinName_Setter_should_sets_font_for_the_latin_characters_only_for_the_specified_shape`
2. Add this at the end: `this.SaveResult(pres);`
3. Open `test\TestResults\LatinName_Setter_should_sets_font_for_the_latin_characters_only_for_the_specified_shape\result.pptx` in PowerPoint

Result

Shows dialog with this message: "PowerPoint found a problem with content in (this file). PowerPoint can attempt to repair the presentation."

Expected

Presentation should load without errors.

Issue #828, opened 2024-12-24
Setting font latin name on a shape also changes the font latin name on other shapes


jcoliz:
Updating the font in one shape also updates the font in another shape.

Use test [LatinName_Setter_should_sets_font_for_the_latin_characters_only_for_the_specified_shape](https://github.com/ShapeCrawler/ShapeCrawler/blob/a5b0639ca791945496283b562fa7f90336ba3c48/test/ShapeCrawler.Tests.Unit/FontTests.cs#L315) to reproduce this bug.

---

### Initial Message

Following test fails

```c#
public void LatinName_Setter_only_sets_current_shape()
{
    // Arrange
    var pres = new Presentation();
    var shapes = pres.Slides[0].Shapes;
    shapes.AddShape(10,20,30,40);
    var shape1 = shapes.Last();
    shapes.AddShape(100,20,30,40);
    var shape2 = shapes.Last();
    shape1.TextBox!.Text = "Shape 1";
    shape1.TextBox!.Paragraphs[0].Portions[0].Font.LatinName = "Segoe UI Semibold";

    // Act
    shape2.TextBox!.Text = "Shape 2";
    shape2.TextBox!.Paragraphs[0].Portions[0].Font.LatinName = "Aptos";

    // Assert
    shape1.TextBox!.Paragraphs[0].Portions[0].Font.LatinName.Should().Be("Segoe UI Semibold");
}
```

It's like they all share the same font?

PR #863, opened 2025-01-07
Fix powerpoint not load correctly (Issue #828)

FabrisTB:
The reason it fails to load the powerpoint correctly was because another LatinName was created in the text RunProperty. Also, some text don't have RunProperties bur EndParagraphRunProperties, we should use it instead.

---

## PR-Codex overview
This PR focuses on improving the handling of strings, color names, and font properties in the codebase. It enhances the robustness of string comparisons, refines font property assignments, and ensures proper initialization of font-related objects.

### Detailed summary
- Modified `StringOf` method in `Assets.cs` to read a file's content.
- Changed string comparison in `HexFromName` method from `ToLower()` to `Equals()` with `StringComparison.CurrentCultureIgnoreCase`.
- Improved font property handling in `TextPortionFont.cs` by checking for null values and appending `LatinFont` correctly.

### DIFF
--- a/src/ShapeCrawler/Shared/Assets.cs
+++ b/src/ShapeCrawler/Shared/Assets.cs
@@ -11,7 +11,7 @@ internal readonly ref struct Assets
     
     internal MemoryStream StreamOf(string file)
     {
-        var stream = this.assembly.GetManifestResourceStream($"ShapeCrawler.Resources.{file}") !;
+        using var stream = this.assembly.GetManifestResourceStream($"ShapeCrawler.Resources.{file}") !;
         var asset = new MemoryStream();
         stream.CopyTo(asset);
         asset.Position = 0;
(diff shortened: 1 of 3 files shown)


src/ShapeCrawler/Shared/Assets.cs
```diff
@@ -11,7 +11,7 @@ internal readonly ref struct Assets
     
     internal MemoryStream StreamOf(string file)
     {
-        var stream = this.assembly.GetManifestResourceStream($"ShapeCrawler.Resources.{file}") !;
+        using var stream = this.assembly.GetManifestResourceStream($"ShapeCrawler.Resources.{file}") !;
```
ashahabov, line 14: `using` statement doesn't make sense here because this stream is a _managed_ resource, i.e., in compile time we know that we do not deal with _unmanaged_ resources.

jcoliz:
LGTM. Thanks very much for this.

Fixes #842

src/ShapeCrawler/Shared/Assets.cs
FabrisTB: I delete it. I remembered reading somewhere the stream should be closed and when I used it before, I always close it. Now that I think about it, it was always on ressources I use only once.

PR #880, opened 2025-01-11
Improve error for non image formats

### LINKED ISSUE #838: AddPicture() should throw SCException for non-image stream
Currently, the below test throws an `System.Xml.XmlException` with message 'Data at the root level is invalid. Line 1, position 1.' Better would be to catch this and re-throw an SCException explaining the conceptual problem which is that you seem to have supplied a stream which cannot be loaded as an image.

```c#
[Test]
public void AddPicture_throws_exception_when_the_specified_stream_is_non_image()
{
    // Arrange
    var pres = new Presentation();
    var shapes = pres.Slide(1).Shapes;
    var stream = StreamOf("autoshape-case011_save-as-png.pptx");

    // Act
    var addingPicture = () => shapes.AddPicture(stream);

    // Assert
    addingPicture.Should().Throw<Exception>();
}
```

blockingHD:
Moved all image processing to Magick.Net in the add picture function. This allows us to attempt to convert unsupported images to png. PowerPoint converts to PNG or JPEG depending on the file format from what I can tell. It probably checks if the format has an Alpha channel.

I have added an array of supported image formats based on my testing. I am not sure if there is a master list for this? 

As for vector formats, Magick.net doesn't support vector-to-vector conversions, so that isn't viable in this case. Hence, SVG is the only supported vector format; emf also works, but this is a mixed format.

Fixes #838

---

## PR-Codex overview
This PR focuses on enhancing image handling in the `ShapeCrawler` library by removing the `Mime` property from `ImageStream`, introducing `EnvironmentChecks` for Ghostscript presence, and updating tests to support various image formats.

### Detailed summary
- Removed the `Mime` property from `ImageStream`.
- Added `EnvironmentChecks` class with method to check Ghostscript installation.
- Updated `ShapeCrawler.Tests.Unit` to include tests for various image formats.
- Modified `SlideShapes` to use `MagickImage` for image processing.
- Enhanced error handling for unsupported image formats.
- Updated test cases to validate image addition with various formats.


src/ShapeCrawler/ShapeCollection/SlideShapes.cs
```diff
@@ -121,68 +123,66 @@ public void AddAudio(int x, int y, Stream audio, AudioType type)
...
+            var cyEmu = UnitConverter.VerticalPixelToEmu(height);
+            var transform2D = pPicture.ShapeProperties!.Transform2D!;
+            transform2D.Offset!.X = xEmu;
+            transform2D.Offset!.Y = yEmu;
+            transform2D.Extents!.Cx = cxEmu;
+            transform2D.Extents!.Cy = cyEmu;
+        }
+        catch (MagickException)
+        {
+            throw new SCException("Unsupported image format.");
```
blockingHD, line 184: This should be "x is an unsupported image type."

src/ShapeCrawler/ShapeCollection/SlideShapes.cs
```diff
@@ -121,68 +123,66 @@ public void AddAudio(int x, int y, Stream audio, AudioType type)
...
+
+            if (imageMagick.Format == MagickFormat.Unknown)
+            {
+                throw new SCException("Unsupported image format.");
+            }
+
+            var originalFormat = imageMagick.Format;
+            if (!SupportedImageFormats.Contains(imageMagick.Format) || VectorImageFormats.Contains(imageMagick.Format))
+            {
+                imageMagick.Format = MagickFormat.Png;
```
blockingHD, line 140: Should this be an option that's passed into the function? Or maybe a global setting?

src/ShapeCrawler/ShapeCollection/SlideShapes.cs
```diff
@@ -121,68 +123,66 @@ public void AddAudio(int x, int y, Stream audio, AudioType type)
...
-            height = imageMagick.Height < 500 ? (int)imageMagick.Height : 500;
 
             if (width == 500 || height == 500)
             {
-                imageMagick.Resize((uint)width, (uint)height);
+                imageMagick.Resize(width, height);
             }
-            
+
+            P.Picture pPicture;
```
blockingHD, line 163: Combine declaration and assignment.

ashahabov:
@blockingHD, please back-merge `master`. I've ignored flaky tests there.

src/ShapeCrawler/ShapeCollection/SlideShapes.cs
blockingHD: @ashahabov @jcoliz I need some input on this, and then I think this is ready.

ashahabov:
> I have added an array of supported image formats based on my testing. I am not sure if there is a master list for this?

No, there is not yet a certain list of supported image formats.

jcoliz:
@ashahabov There's a lot going on here. Please give me a few days and I will run a couple tests.

My first thought is that we for sure should have a unit test for each kind of image we can convert, not only PSD.

We could be a little more thorough in the conversion test. We could save out the presentation, load in as XML, find the image part, and do an image compare against known PNG asset we've already converted and added statically. Possibly room for future improvement.

jcoliz:
We also need to set the mimetype correctly, and should check in unit test. As I understand this PR today a TIFF would get added as a TIFF but would end up with an image/png mimetype, thanks to `SlideShapes.Mime()`

src/ShapeCrawler/ShapeCollection/SlideShapes.cs
jcoliz: @blockingHD you said PowerPoint converts unsupported types to png "from what I can tell". We should be able to discern this precisely. When saving PPTX from PowerPoint, then unzipping, what is the file in the media directory?

@ashahabov In this situation, the question is: To what image file format do we convert image files of other supported formats? Once we know exactly what PowerPoint does, IMO it's fine to hard-code that.

ashahabov: > To what image file format do we convert image files of other supported formats?

I've tested adding on slide five below image formats using PowerPoint:
`.jpg` -> `.jpg`
`.png` -> `.png`
`.gif` -> `.gif`
`.bmp` -> **`.png`**
`.svg` -> **`.png`** (rasterized version)

As @blockingHD has already mentioned, it looks like PowerPoint converts "unsupported" formats in `.png`. Thus, if we can, we should do the same.

jcoliz: Ok, that's interesting. Subtle thing in that list. `.bmp` -> `.png`. This tells us that BMP is _not_ a "supported format", because it is converted by PowerPoint. Ergo `MagickFormat.Bmp` doesn't belong in `SupportedImageFormats` field.

This does really show the need for a test on each "supported format" to make sure it's _not_ converted.

blockingHD: Hi both,

I converted an image to all possible image formats that I could using image magik then took the file types that saved from the media directory. If bmp converts to png then I may have generated a bad file or something. I'll check this out tomorrow.

blockingHD: I'll also add a test for each image type added in.

blockingHD:
I've read all the feedback so far. I'll add the extra tests and do some more testing of the PowerPoint behaviour this week. Thanks for the initial review both :)

jcoliz:
@blockingHD First off, I tested the latest of this against my image collection, and all is well. In fact it shaved off 0.6MB from the 11MB size of one image-heavy file. My images are mostly SVG with < 5% PNGs, and no other file types.

One thing I found. PowerPoint _does_ have a list of supported input file types. Insert > Pictures > This Device, then, there is a drop-down just above open/cancel in the lower right. Expand that, and there is a list of file types. I suggest that this PR should have a unit test with all of those file types at least.

Further, what I found is that some of those are converted (e.g. WebP and BMP were converted to PNG, HEIC converted to JPG), and others were imported as-is (e.g. TIFF, GIF).

Also I found that any image file which wasn't on that list, PowerPoint refused to import. (e.g. PBM, TGA). I'd argue we can overachieve and import anything ImageMagick can handle.

ashahabov:
I have a little different suggestion. Let's take this list as a base and not try to add support for the whole list in this one PR. We can do that later (separate issue). If the adding image is not from this list (or potentially can be handled by Magick.NET), we will throw an exception, e.g. “The stream is not an image or a non-supported image format. You can raise a discussion at https://github.com/ShapeCrawler/ShapeCrawler/discussions to find out about the possibilities supporting it".

- [ ] Windows Enhanced Metafile (_.emf_)
- [ ] Windows Metafile (_.wmf_)
- [ ] Format JPEG (_.jpg or .jpeg_)
- [ ] Format PNG (_.png_)
- [ ] Windows Bitmap (_.bmp_)
- [ ] Format GIT (_.gif_)
- [ ] Compressed Windows Enhanced Metafile (_.emz_)
- [ ] Compressed Windows Metafile (_.wmz_)
- [ ] Tag Image File Format (_.tif_ or _.tiff_)
- [ ] Scalable Vector Graphics (_.svg_)
- [ ] Icon (_.ico_)
- [ ] AV1 Image File Format (_.avif_)
- [ ] WebP (._webp_)

blockingHD:
There has been alot done now so here is how I have addressed each comment above: 

> We also need to set the mimetype correctly, and should confirm this in the unit test. As I understand this PR today a TIFF would get added as a TIFF but would end up with an image/png mimetype, thanks to `SlideShapes.Mime()`
> 
> I think it would be reasonable to pass in Content Type into CreatePPicture, because you'll likely have it handy in AddPicture, where the image is resident in memory. When AddAudio calls it, it can hard-code image/png at that point.

We are now using MagickFormatInfo to grab the Mime type which is then passed into CreatePPicture.

> First off, I tested the latest of this against my image collection, and all is well. In fact it shaved off 0.6MB from the 11MB size of one image-heavy file. My images are mostly SVG with < 5% PNGs, and no other file types.

Using .Strip() removes all the EXIF data etc of the image which both reduces the size of the file and resolves the issue of the flakey tests. This was done over the current implementation as that seemed to blow up the file size by 4-5x.

> Further, what I found is that some of those are converted (e.g. WebP and BMP were converted to PNG, HEIC converted to JPG), and others were imported as-is (e.g. TIFF, GIF).

I have implemented the conversion to PNG vs JPEG by checking if the original image has an alpha channel. This is likely what PowerPoint is doing under the hood, but regardless, doing this check means we aren't adding an Alpha channel where it isn't needed keep file sizes down. 

To Both:
> Let's take this list as a base and not try to add support for the whole list in this one PR.

> Also I found that any image file which wasn't on that list, PowerPoint refused to import. (e.g. PBM, TGA). I'd argue we can overachieve and import anything ImageMagick can handle.

Magick.Net will decode the image in the read step `new MagickImage(...)`, which, if unsupported, throws a MagickDelegateErrorException or similar, which is caught and then used to return the suggested error. So if the library supports it we will convert otherwise throw an error.

> [ ] Icon (_.ico_)
ICO files are supported for reads as listed [here](https://imagemagick.org/script/formats.php) however errors so I have ignored the test for now.

test/ShapeCrawler.Tests.Unit/Helpers/EnvironmentChecks.cs
```diff
@@ -0,0 +1,30 @@
+namespace ShapeCrawler.Tests.Unit.Helpers;
+
+public static class EnvironmentChecks
+{
+    public static bool IsGhostscriptInstalled()
```
blockingHD, line 5: EPS images are supported if the source system has ghostscript so this checks if it is installed. Later the test checks this and marks it as inconclusive if it is. This allows us to check another of the possible exceptions thrown by the library.

ashahabov: I think we should handle this differently, in the production code:

```C#
AddPicture(Stream image)
//...
catch (MagickDelegateErrorException ex) when (ex.Message.StartsWith("FailedToExecuteCommand `\"gswin64c.exe\""))
{
    throw new SCException(
        "It seems the stream is the EPS file format that requires gswin64c.exe which is not installed on your system.");
}
//...
```

ashahabov (changes requested):

test/ShapeCrawler.Tests.Unit/Helpers/EnvironmentChecks.cs
ashahabov: I think we should handle this differently, in the production code:

```C#
AddPicture(Stream image)
//...
catch (MagickDelegateErrorException ex) when (ex.Message.StartsWith("FailedToExecuteCommand `\"gswin64c.exe\""))
{
    throw new SCException(
        "It seems the stream is the EPS file format that requires gswin64c.exe which is not installed on your system.");
}
//...
```

blockingHD: Sorry, this is one example of a format supported if GhostScript is installed. The whole list can be found here https://imagemagick.org/script/formats.php. I wonder if we should just wrap the Magick.Net Exceptions as they are quite descriptive themselves?

ashahabov: > this is one example of a format supported if GhostScript is installed.

Then the message can be more general, e.g. "It seems the stream is an image format that requires GhostScript which is not installed on your system."

> I wonder if we should just wrap the Magick.Net Exceptions as they are quite descriptive themselves?

Currently, it is a convention that caught exceptions must be wrapped in `SCException`. One of the reasons is that the client application developer can more quickly discover that the problem is in the ShapeCrawler, and we _maybe_ can help because we at the compile time understood the source problem.

You can use constructor `SCException(string message, Exception innerException)` to additionally pass Magick.Net Exception.

blockingHD: Done. The check has to be broader as there are differences between the Ghostscript errors depending on the environment.

PR #885, opened 2025-01-13
Remove timestamps from generated PNGs

### LINKED ISSUE #883: Fix intermittent failures in tests due to ImageMagik Timestamp issue
The following tests are removed from CI runs for causing intermittent failures.

```
public void AddPicture_should_not_duplicate_the_image_source_When_the_same_svg_image_is_added_twice()
public void AddPicture_should_not_duplicate_the_image_source_When_the_same_svg_image_is_added_on_two_different_slides()
public void AddPicture_should_not_duplicate_the_image_source_When_the_same_svg_image_is_added_to_a_loaded_presentation()
```

Because these test functionality that's on the critical path for me, I'd like to work on getting them back. Unfortunately, they only seem to fail on CI builds. I have never gotten any local failure on these, so it's hard to debug.

I'll work on setting up a github workflow in my own fork to see if I can add enough information in here to get more info.

jcoliz:
This fixes a bug introduced in PR #830. When adding SVGs, the rasterized PNG was being created with a timestamp. This interfered with the logic to de-duplicate image assets in a single presentation.

@ashahabov Couple things to consider. Adding delay to these tests was required to make them reliably fail. Leaving it in will increase test execution time. Second, I added categories on these test to make them easy to run repeatedly. Could remove this if you consider it clutter.

Fixes #883

---

## PR-Codex overview
This PR updates image handling in the `ShapeCollection` and modifies test attributes for better categorization.

### Detailed summary
- In `SlideShapes.cs`, changed the `Write` method call on `imageMagick` to include `PngWriteDefines` with `IncludeChunks` set to `PngChunkFlags.None`.
- In `ShapeCollectionTests.cs`, replaced `[Explicit]` attributes with `[Category("issue-883")]` for three test methods to categorize them under a specific issue.

### DIFF
--- a/test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
+++ b/test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
@@ -350,7 +350,6 @@ public class ShapeCollectionTests : SCTest
     }
 
     [Test]
-    [Explicit("Fails. Tracked in issue #883")]
     [Category("issue-883")]
     public async Task AddPicture_should_not_duplicate_the_image_source_When_the_same_svg_image_is_added_twice()
     {
@@ -361,7 +360,7 @@ public class ShapeCollectionTests : SCTest
 
         // Act
         shapes.AddPicture(svgImage);
-        Task.Delay(1100).Wait();
+        Task.Delay(1100).Wait(); // See issue #883
         shapes.AddPicture(svgImage);
 
         // Assert
@@ -372,7 +371,6 @@ public class ShapeCollectionTests : SCTest
     }
     
     [Test]
-    [Explicit("Fails. Tracked in issue #883")]
     [Category("issue-883")]
     public void AddPicture_should_not_duplicate_the_image_source_When_the_same_svg_image_is_added_on_two_different_slides()
     {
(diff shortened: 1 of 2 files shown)


test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
```diff
@@ -360,6 +360,7 @@ public void AddPicture_should_not_duplicate_the_image_source_When_the_same_svg_i
 
         // Act
         shapes.AddPicture(svgImage);
+        Task.Delay(1100).Wait(); // See issue #883
```
ashahabov, line 363: I think using this statement in Act-block turns this test into a non-functional test. I consider unit tests more as functional/component tests, and Act-block contains only code that the library's client uses.

Maybe we should start moving such tests into a separate test project like "ShapeCrawler.Stress(Load)Tests". The tests in this project should not be run in parallel and should be run multiple times.

I do not have a lot of experience in load testing, and therefore I'm open to ideas.

jcoliz: No worries. I have removed them. 952140244bc4e23144eff556e69eed2da53f73ea

PR #893, opened 2025-01-19
Add column duplication

### LINKED ISSUE #661: Add columns to table
Add the capability to add columns to the table.

---

### Discussed in https://github.com/ShapeCrawler/ShapeCrawler/discussions/650

<div type='discussions-op-text'>

<sup>Originally posted by **DorijanH** April  5, 2024</sup>
Hi there,

is there a way to manipulate the columns in a table, for example adding a new column or duplicating an existing one? Something similar to the `Duplicate` method available in the `ITableRow` interface</div>

blockingHD:
Looking to resolve #661 this PR adds a duplicate method to the column class. Alongside this it brings the Columns implementation inline with the Rows implementation on the Table object.

---

## PR-Codex overview
This PR focuses on enhancing the `ITableColumns` interface and its implementation, introducing a `Duplicate` method for columns, and updating related tests to ensure proper functionality.

### Detailed summary
- Added `Duplicate` method to the `IColumn` interface and implemented it in the `Column` class.
- Introduced `ITableColumns` interface for managing table columns.
- Updated `ITable` to use `ITableColumns`.
- Removed obsolete methods from `ITable`.
- Enhanced tests for column duplication and proportional width adjustment.


test/ShapeCrawler.Tests.Unit/TableTests.cs
```diff
@@ -90,14 +91,72 @@ public void InsertColumnAfter_inserts_column_after_the_specified_column_number()
...
+    {
+        // Arrange
+        var pptx = TestAsset("table-case001.pptx");
+        var pres = new Presentation(pptx);
+        var table = pres.Slides[0].Shapes.GetByName<ITable>("Table 1");
+        var column = table.Columns[0];
+        var columnsCountBefore = table.Columns.Count;
+
+        // Act
+        column.Duplicate();
```
ashahabov, line 113: For example, in this case, when the table has three columns and the first is selected. `.Duplicate()` adds duplication at the end of the table as the fourth column, but my expectation was that it adds the duplicated column right after being selected as the second column.

I think most users will also expect such behavior. @blockingHD, what do you think about this?

blockingHD: Funnily enough, I implemented it this way to begin with then decided to implement it inline with the duplicate method on the TableRow! Happy to go this way though.

ashahabov: Then we should replace the summary of `IColumn.Duplicate` with, e.g., `Adds a copy of the current column at the end of the table.`

ashahabov:
LGTM. @blockingHD, is it ready to merge?

blockingHD:
@ashahabov sorry I miss clicked the review button. However, this is now in line with the rows duplicate method. I was just looking to fix the fact that it will currently respread the columns evenly throughout the table's width. I have worked out that PowerPoint reduces all the columns proportionally so I have just implemented that. I will push it now but I need to write unit tests and double check the logic.

blockingHD:
@ashahabov This now works and will size the columns proportionally, keeping the same width of the table. I wonder if the same logic should also be moved over to the add column logic, as that currently redistributes the columns evenly. However, that should probably be a subsequent PR.

Issue #808, opened 2024-12-15
Support adjustments for various geometry shapes


jcoliz:
Many preset geometry shapes allow adjustment in the UI. This is shown by a small yellow circular anchor which the user can drag. Currently, ShapeCrawler allows adjustment of Rounded Rectangles and only the top of Top Corners Rounded Rectangle. Yet there are many other possible adjustments to be made to shapes.

Current thought is that `ShapeGeometry.CornerSize` is too narrowly-scoped. Perhaps better would be a property which gets/sets an array of adjustments.

Alternately, we could write a separately-named method for each kind of adjustment to each kind of adjustable shape, as done today with `CornerSize`.

ashahabov:
I'm more in favor of the alternative option, but am open to a discussion.

### `AdjustmentCollection` vs Specific Property

```C#
var pres = new Presentation("pres.pptx");
var shapeCollection = pres.Slide(1).Shapes;

var adjustmentCollection = shapeCollection.GetByName("Rectangle").Adjustments;
// adjustmentCollection is null because of Rectangle shape is not adjustable

adjustmentCollection = shapeCollection.GetByName("Rectangle: Rounded Corners").Adjustments;
var adjustment0 = adjustmentCollection[0];
// adjustment0.Name == "Corner"
adjustment0.Value = 10;

adjustmentCollection = shapeCollection.GetByName("Rectangle: Top Corners Snipped").Adjustments;
adjustment0 = adjustmentCollection[0];
// adjustment0.Name == "Top Snip"
adjustment0.Value = 10;
var adjustment1 = adjustmentCollection[1];
// adjustment1.Name == "Bottom Snip"
adjustment1.Value = 20;
```

```C#
var pres = new Presentation("pres.pptx");
var shapeCollection = pres.Slide(1).Shapes;

var roundedCornersRectangle = shapeCollection.GetByName<IRoundedCornersRectangle>("Rectangle: Rounded Corners");
roundedCornersRectangle.Corner = 10;

var topCornersSnippedRectangle = shapeCollection.GetByName<ITopCornersSnippedRectangle>("Rectangle: Top Corners Snipped");
topCornersSnippedRectangle.TopSnip = 10;
topCornersSnippedRectangle.BottomSnip = 20;
```

jcoliz:
I was thinking more like this for first option:

```c#
var pres = new Presentation("pres.pptx");
var shapeCollection = pres.Slide(1).Shapes;

[...]

PR #898, opened 2025-01-23
Adds an internal Adjustments property for geometry shapes with adjustments

jcoliz:
Adds a getter for adjustments which works for all kinds of geometries.

Adds a setter for adjustments which works only for rectangles. Can be extended to any geometries by adding to the mapping table.

Originally was thinking this would be internal only, and then exposed as specific properties a'la "Corner Size". But it will take a significant effort to work out the exact meanings of all shapes' adjustments. So now thinking it would still be useful to make this public

Starting #808

---

## PR-Codex overview
This PR introduces the `Adjustments` property across multiple shape classes, allowing for geometry-specific adjustments. It also updates the handling of adjustments in the `ShapeGeometry` class and enhances testing for these adjustments.

### Detailed summary
- Added `Adjustments` property in `Shape`, `GroupedShape`, `Picture`, `RootShape`, and `AutoShape` classes.
- Implemented logic to manage adjustments in `ShapeGeometry`.
- Updated tests in `ShapeTests` to validate adjustments functionality.
- Enhanced error handling for adjustment limits.


test/ShapeCrawler.Tests.Unit/ShapeTests.cs
```diff
@@ -765,4 +767,129 @@ public void Name_Setter_sets_grouped_shape_name()
...
+
+    public void Adjustments_getter_returns_values(string name, string expectedAdjustmentsJson)
+    {
+        // Arrange
+        var shape = 
+            new Presentation(TestAsset("062_shape-adjustments.pptx"))
+                .Slides[0]
+                .Shapes
+                .GetByName(name)
+                as RootShape;
```
ashahabov, line 861: `RootShape` is an `internal` class. Internals are exposed to the test project for exclusive situations, especially for use in Assert block.

Is there any way to avoid using `RootShape` in test?

jcoliz: Yah. The solution will be to make it part of the public interface, right? I had originally hesitated on that, hoping to have a better public interface. But by adding this to `IGeometry`, then any `IShape` will do in this context.

ashahabov: > The solution will be to make it part of the public interface, right?

Right.

jcoliz: Ok, that's done in f9913d377ac3ce265314bd87575f65b81b008e09

src/ShapeCrawler/ShapeCollection/ShapeGeometry.cs
```diff
@@ -105,33 +131,53 @@ public decimal CornerSize
...
-            }
-
-            if (shapeType == A.ShapeTypeValues.Round2SameRectangle)
-            {
-                return this.ExtractCornerSizeFromRound2SameRectangle();
-            }
-
-            return 0;
+                (Geometry.RoundedRectangle, 0) or 
+                (Geometry.TopCornersRoundedRectangle, 0) => 35,
```
ashahabov, line 138: If `35` is the defaultze, then please add comment like `// default corner size predefined in PowerPoint` to avoid being it a magic number.

jcoliz: 0dbde615f47d0052cd67d33db03534162af53760

PR #936, opened 2025-02-15
Remove tIME chunk from generated PNGs

### LINKED ISSUE #883: Fix intermittent failures in tests due to ImageMagik Timestamp issue

blockingHD:
Removed tIME chunk from generated PNG to fix #883 (again!) this keep all tests passing.

I would use the PngChunkFlags enum, however, it does not include tIME flag, so it has to be this hard-coded string.

---

## PR-Codex overview
This PR focuses on enhancing image handling in the `ShapeCrawler` library by ensuring that adding the same image multiple times does not create duplicates in presentations. It introduces a setting to exclude certain PNG chunks and updates existing tests for better reliability.

### Detailed summary
- Added `imageMagick.Settings.SetDefine("png:exclude-chunk", "tIME")` in `SlideShapes.cs`.
- Removed multiple test cases for image duplication in `ShapeCollectionTests.cs`.
- Introduced a new test case for image duplication with a delay in `ShapeCollectionTests.cs`.
- Updated `AddPicture_should_not_duplicate_the_image_source_When_the_same_svg_image_is_added_to_a_loaded_presentation` to include a delay.
- Ensured that the same image added twice results in only one instance in the presentation.

### DIFF
--- a/test/ShapeCrawler.Tests.Load/ShapeCollectionTests.cs
+++ b/test/ShapeCrawler.Tests.Load/ShapeCollectionTests.cs
@@ -6,8 +6,6 @@ namespace ShapeCrawler.Tests.Load;
 public class ShapeCollectionTests : SCTest
 {
     [Test]
-    [Repeat(40)]
-    [Explicit("Flaky test. Read https://github.com/ShapeCrawler/ShapeCrawler/issues/883")]
     public void AddPicture_should_not_duplicate_the_image_source_When_the_same_svg_image_is_added_to_a_loaded_presentation()
     {
         // Arrange
@@ -19,6 +17,7 @@ public class ShapeCollectionTests : SCTest
         var loadedPres = SaveAndOpenPresentation(pres);
 
         // Act
+        Thread.Sleep(1000);
         shapes = loadedPres.Slides[0].Shapes;
         shapes.AddPicture(image);
 
@@ -32,8 +31,6 @@ public class ShapeCollectionTests : SCTest
     }
     
     [Test]
-    [Repeat(40)]
-    [Explicit("Flaky test. Read https://github.com/ShapeCrawler/ShapeCrawler/issues/883")]
     public void AddPicture_should_not_duplicate_the_image_source_When_the_same_svg_image_is_added_twice()
     {
         // Arrange
@@ -43,6 +40,7 @@ public class ShapeCollectionTests : SCTest
 
         // Act
         shapes.AddPicture(svgImage);
+        Thread.Sleep(1000);
         shapes.AddPicture(svgImage);
 
         // Assert
--- a/test/ShapeCrawler.Tests.Load/ShapeCollectionTests.cs
+++ b/test/ShapeCrawler.Tests.Load/ShapeCollectionTests.cs
@@ -6,7 +6,8 @@ namespace ShapeCrawler.Tests.Load;
 public class ShapeCollectionTests : SCTest
 {
     [Test]
-    public async Task AddPicture_should_not_duplicate_the_image_source_When_the_same_svg_image_is_added_to_a_loaded_presentation()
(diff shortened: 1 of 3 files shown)


jcoliz:
> Introduced await Task.Delay(1000) in both test methods to handle potential timing issues during image addition.

@blockingHD So while I agree with this change, I seem to recall that @ashahabov didn't want this, last time we went through the previous round of investigation on the flaky tests.

blockingHD:
@jcoliz #896 added the delay in the end but there are separate tests for it which still exist however SVG wasn't included in that list which is why this was missed when I did the last change. I will revert the test change and add svg to that test suit instead.

ashahabov:
> @blockingHD So while I agree with this change, I seem to recall that @ashahabov didn't want this, last time we went through the previous round of investigation on the flaky tests.

Right. However, as suggested [then ](https://github.com/ShapeCrawler/ShapeCrawler/pull/885#discussion_r1913705675)I have started moving such tests to the separate test [ShapeCrawler.Tests.Load](https://github.com/ShapeCrawler/ShapeCrawler/tree/master/test/ShapeCrawler.Tests.Load).

test/ShapeCrawler.Tests.Unit/ShapeCollectionTests.cs
```diff
@@ -419,6 +419,29 @@ public void AddPicture_should_not_duplicate_the_image_source_When_the_same_image
         imageParts.Count.Should().Be(1);
     }
     
+    [Test]
+    public void AddPicture_should_not_duplicate_the_png_image_source_When_the_same_svg_image_is_added_a_second_apart()
```
ashahabov, line 423: Please move this test into the test project `ShapeCrawler.Tests.Load`. Here is [Why](https://github.com/ShapeCrawler/ShapeCrawler/pull/885#discussion_r1913705675).

blockingHD: I've moved both the svg and raster tests across.

test/ShapeCrawler.Tests.Load/ShapeCollectionTests.cs
```diff
@@ -7,7 +7,6 @@ public class ShapeCollectionTests : SCTest
 {
     [Test]
     [Repeat(40)]
-    [Explicit("Flaky test. Read https://github.com/ShapeCrawler/ShapeCrawler/issues/883")]
     public void AddPicture_should_not_duplicate_the_image_source_When_the_same_svg_image_is_added_to_a_loaded_presentation()
```
ashahabov, line 10: As @jcoliz mentioned [here](https://github.com/ShapeCrawler/ShapeCrawler/issues/883#issuecomment-2661039496), then do we need in this test and in `AddPicture_should_not_duplicate_the_image_source_When_the_same_svg_image_is_added_twice` also use `Task.Delay(1000).Wait()` instead of the atrribute `Repeat(40)`?

blockingHD: I have merged the two normal svg tests and replaced the repeats with Task.Delay

Issue #552, opened 2023-07-09
Table height is not updating


ashahabov:
Increasing the table height doesn't increase its row height.

### Steps to Reproduce

Use test [Height_Setter_should_proportionally_increase_the_row_heights_When_the_new_table_height_is_bigger](https://github.com/ShapeCrawler/ShapeCrawler/blob/1cc3b4bc0375cd888b5f88b6f1b7c7e29916f10e/test/ShapeCrawler.Tests.Unit/TableTests.cs#L1063-L1079) to reproduce this bug:

```C#
[Test]
public void Height_Setter_should_proportionally_increase_the_row_heights_When_the_new_table_height_is_bigger
{
    // Arrange
    var pres = new Presentation();
    pres.Slide(1).Shapes.AddTable(10, 10, 2, 2);
    var addedTable = pres.Slide(1).Shapes.Last<ITable>();
    
    // Act
    addedTable.Height = 100;
    
    // Assert
    addedTable.Rows[0].Height.Should().Be(40);
    addedTable.Rows[1].Height.Should().Be(40);
    pres.Validate();
}
```

FabrisTB:
Isn't this already implemented ? When we update table rows height, it automatically change table height.

PR #952, opened 2025-02-27
Fix for issue  #552 : Table height is not updating 

FabrisTB:
Also fix the unit test

---

## PR-Codex overview
This PR focuses on enhancing the `Height` property functionality for tables in the `ShapeCrawler` library. It ensures that when the table height is updated, the row heights are proportionally adjusted.

### Detailed summary
- Removed the `[Explicit]` attribute from the test `Height_Setter_should_proportionally_increase_the_row_heights_When_the_new_table_height_is_bigger`.
- Changed the height assignment to use a proportional increase based on the current row height.
- Added a new `Height` property in `ITable` that overrides the base property and calls `UpdateTableHeight`.
- Introduced `SetTableHeight` method in `ITable` to set the height and update row heights accordingly.
- Added `UpdateTableHeight` method in `ITable` to calculate and set new row heights based on the updated table height.
- Implemented `SetHeight` method in `ITableRow` to adjust the row height in pixels. 
- Updated height adjustments in `UpdateHeight` to call `SetTableHeight` instead of directly modifying the parent table height.

### DIFF
--- a/src/ShapeCrawler/Tables/ITable.cs
+++ b/src/ShapeCrawler/Tables/ITable.cs
@@ -86,6 +86,12 @@ internal sealed class Table : CopyableShape, ITable
         set => this.SetTableStyle(value);
     }
 
+    public new decimal Height 
+    { 
+        get => base.Height;
+        set => this.UpdateTableHeight(value); 
+    }
+
     public ITableStyleOptions TableStyleOptions { get; }
 
     public override bool Removeable => true;
@@ -144,6 +150,11 @@ internal sealed class Table : CopyableShape, ITable
 
     public override ITable AsTable() => this;
 
+    internal void SetTableHeight(decimal value)
+    {
+        base.Height = value;
+    }
+
     private void SetTableStyle(ITableStyle style)
     {
         this.ATable.TableProperties!.GetFirstChild<A.TableStyleId>() !.Text = ((TableStyle)style).Guid;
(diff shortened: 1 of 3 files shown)


FabrisTB:
While making this commit, I saw that in some place you use points in others you use pixels. Wouldn't it be better if we use either pixel or point everywhere and not mixing them ?

ashahabov:
> While making this commit, I saw that in some place you use points in others you use pixels. Wouldn't it be better if we use either pixel or point everywhere and not mixing them ?

@FabrisTB, thank you for the PR. Sorry for the delayed review. Have deadline on my main job project.
___

When it is possible, the public interface members should expose measure units as PowerPoint uses. For some of the internal calculations, e.g., in graphics, I think it is ok making a mix.

I believe we use pixels only for graphics, isn't it?

ashahabov:
I've checked what PowerPoint uses: inch🤔

image

Hmm, then probably you are right, pixels are better suited for table sizes. Using inches would be kind of weird. I haven't met in such libraries using inch for this.

FabrisTB:
Using inch is madness, sometimes iirc, they use cm too. As long as it's documented, using either pixels or point is ok I think

ashahabov:
I made today some additional research, and I now think we need for sizes and positions to use **points** to be resolution-independently.

FabrisTB:
so we'll use point then.

PR #961, opened 2025-03-18
Remove newlines when replacing text

### LINKED ISSUE #960: Editing text containing a newline causes newline to be replicated exponentially each time
We're attempting to use the project for the purpose of populating a template powerpoint, e.g put placeholders in a template powerpoint, and then use ShapeCrawler to replace those placeholders.

This works great, aslong as the source text does not contain 'new line' \n characters. If it contains new lines, what happens is that something within ShapeCrawler is doubling the newline characters.

I've created a sample project (attached) that replicates this issue.

In the sample program we have two identical slides:

Image

The sample program uses a replace on this text as follows:

For slide 1, at a top level.

`
var Shape1 = _presentation.Slides.First(x => x.Number == 1).Shapes.First();
Shape1.Text = Shape1.Text.Replace("World", "Earth");
`

For slide 2, at a paragraph level

`
var Shape2 = _presentation.Slides.First(x => x.Number == 2).Shapes.First();
Shape2.TextBox.Paragraphs.First().Text = Shape2.TextBox.Paragraphs.First().Text.Replace("World", "Earth");
`

The resulting slides will look like this:

Image

If you inspect the text after doing the replace, you will notice that ShapeCrawler is doubling the qty of newline characters between Hello and World.

jcoliz:
Previously, when changing the text on a paragraph, the old lines were removed. However, the old NEWLINES were not removed. This PR removes also the old newlines.

Plus a bonus test for `Paragraph.ReplaceText()` for which I could not find test coverage.

Fixes #960

---

## PR-Codex overview
This PR focuses on updating the handling of text in paragraphs within the `ShapeCrawler` library, ensuring that replacing text preserves newline characters. It also modifies test assets and adds unit tests to verify the new behavior.

### Detailed summary
- Removed `Assets\072 content placeholder.pptx` and added `Assets\073 replacing text.pptx` as resources.
- Updated `IParagraph.cs` to modify how remaining runs are handled.
- Added three unit tests in `ParagraphTests.cs` to verify text replacement preserves newlines.


jcoliz:
I suspect that the errors are Linux-specific. Paragraphs code uses `Environment.NewLine`, which is "\n" on Linux. Instead, should use the Windows newline string ("\n\r") directly, which is how Powerpoint UI writes strings to Powerpoint files.

src/ShapeCrawler/Texts/IParagraph.cs
```diff
@@ -95,7 +96,11 @@ public string Text
 
             // To set a paragraph text we use a single portion which is the first paragraph portion.
             var baseARun = this.aParagraph.GetFirstChild<A.Run>() !;
-            var remainingRuns = this.aParagraph.OfType<A.Run>().Where(run => run != baseARun).ToList();
+            var remainingRuns = new List<OpenXmlCompositeElement>();
+
+            remainingRuns.AddRange(this.aParagraph.OfType<A.Run>().Skip(1));
```
ashahabov, line 101: Could you please add a comment for this magic number `1` answering WHY we are skipping here one `Run` element.

jcoliz: @ashahabov Honestly it's not entirely clear why this method always preserves the first Run. Could you explain this? Why not just tear down the aParagraph down to empty, and then rebuild from scratch? In case of this PR, I'm matching the pre-existing logic.

ashahabov: It took me a little while to remember🙂

Actually, the comment is already there, it's the one a bit above — `// To set a paragraph text we use a single portion which is the first paragraph portion.`

We use a `Run` of the older text to preserve the text format, e.g. font style: size, bold, etc.

PR #967, opened 2025-03-22
Correct wrong cast in point

FabrisTB:
I finally found why I didn't use Point class for cell border : point get truncate by the cast when try to convert to other value.


ashahabov:
I've moved these changes to another PR https://github.com/ShapeCrawler/ShapeCrawler/pull/966 where a big refactoring is going on right now.

Issue #892, opened 2025-01-19
ICO images are not supported


blockingHD:
While implementing #880 it was found that ICO is seemingly not supported however this is supported in PowerPoint by converting to PNG. ICO support should be added.

Hasan-75:
@ashahabov I've opened a PR for this issue: #1117.
Feel free to review at your convenience. Thanks!

PR #1117, opened 2025-09-07
Support adding pictures using .ico image

Hasan-75:
This PR resolves issue #892 _(Add picture using .ico image)._

### Description
The problem was occurring in the following function:
```csharp
private static MagickImage CreateMagickImage(Stream imageStream)
{
    return new MagickImage(
        imageStream,
        new MagickReadSettings { BackgroundColor = MagickColors.Transparent
    });
}
```

Currently, Magick.NET attempts to auto-detect the image format from the stream header. While this works reliably for common formats like PNG and JPEG, ICO images were failing when auto-detection could not correctly identify the format. The failure produced the error:
`no decode delegate for this image format ' @ error/blob.c/BlobToImage/481`

### Solution:
The fix is to explicitly pass the image format when creating the MagickImage:
```csharp
private static MagickImage CreateMagickImage(Stream imageStream, MagickFormat format)
{
    return new MagickImage(
        imageStream,
        new MagickReadSettings
        {
            Format = format,
            BackgroundColor = MagickColors.Transparent,
        });
}
```


src/Slides/AudioVideoShapeCollection.cs
```diff
@@ -177,6 +178,8 @@ public void AddVideo(int x, int y, Stream stream)
 
     public void AddPicture(Stream imageStream) => shapes.AddPicture(imageStream);
 
+    public void AddPicture(Stream imageStream, MagickFormat format) => shapes.AddPicture(imageStream, format);
```
AdamShakhabov: It would be good if we still try to determine ICO format by content.

I'm not sure, but LLM suggests using the following method. Please check this one:

```C#
bool IsIco(FileStream stream)
{
    if (stream.Length < 6)
        return false;
    
    var originalPosition = stream.Position;
    stream.Seek(0, SeekOrigin.Begin);
    
    try
    {
        var header = new byte[6];
        var bytesRead = stream.Read(header, 0, 6);
        
        if (bytesRead < 6)
            return false;
        
        // ICO file signature:
        // Bytes 0-1: Reserved (must be 0x00 0x00)
        // Bytes 2-3: Image type (must be 0x01 0x00 for ICO)
        // Bytes 4-5: Number of images (must be >= 1)
        return header[0] == 0x00 && 
               header[1] == 0x00 && 
               header[2] == 0x01 && 
               header[3] == 0x00 &&
               (header[4] > 0 || header[5] > 0); // At least 1 image
    }
    finally
    {
        stream.Seek(originalPosition, SeekOrigin.Begin);
    }
}
```

Hasan-75: Yeah I got the same suggestion from ChatGPT. However I didn't proceed with this approach to keep things as consistent as possible with existing codes.

I can push another commit soon if you think determining image format from stream bytes will be a better approach.

AdamShakhabov: Yes, you can use it. It will be consistent.

Hasan-75: @ashahabov pushed changes as discussed

Issue #945, opened 2025-02-16
Footer text


ashahabov:
Add a capability to add the same footer text on all slides:

![Image](https://github.com/user-attachments/assets/2f1ba3a6-55c6-4cd1-96fd-55e661c37e2b)

### Usage example
```C#
var pres = new Presentation();
pres.Footer.Text = "My footer text";
```

PR #1118, opened 2025-09-11
Added Footer Text Support

Hasan-75:
This PR resolves issue #945 (Footer Text).

### Description:
There are 3 new functions in `IFooter`:
```csharp
    /// <summary>
    ///     Set footer text on all slides.
    /// </summary>
    void AddFooterText(string text);

    /// <summary>
    ///     Removes footer text from all slides.
    /// </summary>
    void RemoveFooterText();

    /// <summary>
    ///     Removes footer shape from all slides.
    /// </summary>
    void RemoveFooter();
```

#### Original Requirement: 
- [X] Allow adding footer text to all slides.

#### Additional Functionalities: 
- [X] Allow clearing footer text from all slides.
- [X] Allow removing footer shapes from all slides.

### Future Scope:
- [ ] Allow showing footers on specific pages.

### DIFF
--- a/src/Presentations/IFooter.cs
+++ b/src/Presentations/IFooter.cs
@@ -110,10 +110,7 @@ internal sealed class Footer(UpdatedSlideCollection slides): IFooter
         foreach (var slide in slides)
         {
             var footerShape = slide.Shapes.FirstOrDefault(shape => shape.PlaceholderType == PlaceholderType.Footer);
-            if (footerShape != null && footerShape.TextBox != null)
-            {
-                footerShape.TextBox.SetText(string.Empty);
-            }
+            footerShape?.TextBox?.SetText(string.Empty);
         }
     }
 
--- a/src/Presentations/IFooter.cs
+++ b/src/Presentations/IFooter.cs
@@ -26,19 +26,14 @@ public interface IFooter
     void RemoveSlideNumber();
 
     /// <summary>
-    ///     Set footer text on all slides.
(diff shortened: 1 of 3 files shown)


src/Presentations/IFooter.cs
```diff
@@ -24,6 +24,21 @@ public interface IFooter
...
+
+    /// <summary>
+    ///     Removes footer text from all slides.
+    /// </summary>
+    void RemoveFooterText();
+
+    /// <summary>
+    ///     Removes footer shape from all slides.
+    /// </summary>
+    void RemoveFooter();
```
AdamShakhabov: What about having only two APIs?
- `IFooter.AddText(string content)` — Adds text footer on all slides with specified content. If a slide already has a text footer, the content will be replaced.
- `IFooter.RemoveText()` — Removes text footers from all slides.

Please let me know if there is a case when we need two separate methods.

Hasan-75: @ashahabov I have no strong reason for having two separate methods to remove footers. I've pushed another commit with changes as per your suggestion. Please take a look.

Also, please let me know what you think about having a functionality to add footer in specific page. I can address that too (if not in this PR, then in a separate one). Should be a fairly easy change.

AdamShakhabov: Yes, adding footer on a certain slide is a good idea! Let's do this separately.

Hasan-75: Created a separate issue #1119. 
Resolving this conversation.

PR #1131, opened 2025-09-29
Update color of chart title

ashahabov:


src/Charts/ChartTitle.cs
```diff
@@ -0,0 +1,243 @@
...
+        if (cChartText == null)
+        {
+            return; // No text content, don't create structure
+        }
+
+        var cRichText = cChartText.GetFirstChild<C.RichText>();
+        if (cRichText == null)
+        {
+            return; // No rich text, don't create structure
+        }
```
Hasan-75: we can combine these:

```csharp

if (cChart?.Title?.GetFirstChild<C.ChartText>()?.GetFirstChild<C.RichText>() is null)
{
    return;
}
```

src/Charts/ChartTitle.cs
```diff
@@ -0,0 +1,226 @@
...
+        if (hex.Length == 8)
+        {
+            hex = hex[..6];
+        }
+
+        // Apply color to all existing runs
+        foreach (var aParagraph in cRichText.Elements<A.Paragraph>())
+        {
+            foreach (var aRun in aParagraph.Elements<A.Run>())
+            {
```
Hasan-75: Instead of nested foreach loops, couldn't we use this?

`foreach (var aRun in cRichText.Descendants<A.Run>())`

AdamShakhabov: We could, but it will be slower because of `cRichText.Descendants<A.Run>()` will search for `A.Run` everywhere not only in the `A.Paragraph` **first level** children.

Hasan-75 (approved):

PR #1241, opened 2025-12-26
Add support for transparent shapes when saving as image

NickKellett:
Currently, a transparent shape renders as solid fill when saving a slide as an image. This PR solves for this and renders transparent shapes and overlays properly in the saved image. 

In the Unit Test, when I load the saved image its Alpha is always 255 when I would expect the transparent value to be something else. I may simply not be understanding how to assess a transparent background loaded from SkiaSharp, or maybe I have somehow setup the wrong test asset. For the moment I have a placeholder assertion that always passes since the code produces the correct effect and I did not want to issue a PR with a failing test.

This PR has a small tweak to avoid null pointer exceptions when a TextBox has a Date object instead of regular text.

### DIFF
--- a/src/Shapes/DrawingShape.cs
+++ b/src/Shapes/DrawingShape.cs
@@ -217,7 +217,7 @@ internal class DrawingShape(Position position, ShapeSize shapeSize, ShapeId shap
         using var fillPaint = new SKPaint();
         
         // Adjust alpha for transparency only if Fill is possible and alpha < 100% (fully opaque/solid)
-        if (this.Fill != null && this.Fill is { Alpha: < 100.0 })
+        if (this.Fill is { Alpha: < 100.0 })
         {
             // Alpha in PowerPoint: 20% transparency means Alpha value is 80
             var alpha = (byte)(255 * (this.Fill.Alpha / 100.0));


ashahabov:
**Build issues**: You should build [Release configuration](https://github.com/ShapeCrawler/ShapeCrawler/blob/master/AGENTS.md#editorconfig-enforcement) to be able to see and resolve _locally_ the current CI failing issues.
**Testing**: Try testing it using snapshot testing ([Verify](https://github.com/VerifyTests/Verify)) how it does it, for example, [SaveImageTo_draws_Text_Shapes](https://github.com/ShapeCrawler/ShapeCrawler/blob/7c03d22f4d1439dd9969b89b840960fa5779bfc3/tests/ShapeCrawler.DevTests/UserSlideTests.cs#L543C5-L587C6). This approach is still experimental, but it should be easier.

NickKellett:
Sorry about that, I've passed the Release configuration locally so hopefully this will be okay.

ashahabov:
@NickKellett , it still fails due to an additional SonarCube check https://github.com/ShapeCrawler/ShapeCrawler/pull/1241/checks?check_run_id=59002988570

PR #1278, opened 2026-03-26
Add functionality to save company name

geskill:
Currently it is not possible to set the company information.
This information is not saved within the `CoreFilePropertiesPart` but `ExtendedFilePropertiesPart`.

I made some small changes to add this functionality, but was not available to make it as clean as in core properties.
Maybe it is better to integrate the company to the other properties. Therefore i mark this as draft.

Usage:
```
presentation.ExtendedProperties.Company = "Test company";
```


ashahabov:
@geskill , thank you for the PR!

I would suggest to extend existing `IPresentationProperties` with `IPresentationProperties.Company`. Then `PresentationProperties`'s constructor should accept additionally `ExtendedFilePropertiesPart`.

ashahabov:
Looks good. We need a test and I think it is done.

